PR #39, opened 2016-11-09
Add VS Code debugging and sourcemaps support

lumaxis:


lumaxis:
@christav I don't remember what we agreed on with regards to this Pull Request. Can you help me jog my memory?

Also, is the `tsc` task actively used? Unfortunately, VS Code currently only allows one command in the `tasks.json` I'd therefore rather use `gulp` and various arguments instead to be more flexible. @kmiecikt ?

christav:
I'm not sure we agreed on anything to be honest. Right now there are merge conflicts anyway, so you'll need to fix those at least.

`tsc` isn't directly used anymore, all that should have been moved over to `gulp`. If you find a direct use of `tsc` outside the `gulpfile` let me know and we'll remove it.

lumaxis:
Sorry, should have been more clear: I was referring to the `tsc` task in the [`tasks.json`](https://github.com/Microsoft/MobileCenter-Cli/blob/d2c56e1716899b71cb49884709d649f785c9c7d6/.vscode/tasks.json). That's also what this merge conflict is about. I couldn't find any place where this is used but @kmiecikt added it so I was waiting to see if he still has a need for it that can't be solved by a gulp task.

christav:
@lumaxis Gotcha. I have no opinion as I don't run/debug in VS Code. Work it out among yourselves. :smile:

PR #67, opened 2016-11-29
Test cloud calabash

kmiecikt:
## Processing `--include` and `--test-parameter` arguments directly in the CLI
All `prepare` commands share two command line arguments: 
1. `--include`. This allows user to add additional files to the Test Cloud upload.
2. `--test-parameter`. This allows user to add additional test parameters to the manifest.

In the current version, these parameters are passed to external `prepare` process (for frameworks that need it). That process parses them, copies to the artifacts directory, and saves appropriate information to the test manifest.
That has two problems:
1. We need to implement parsing and validation multiple times.
2. It is very easy to get inconsistent behavior. For example, the Calabash prepare command separates key / values by using colon (':'), while other frameworks use equal sign ('=').

This PR fixes that. Now the CLI itself is reponsible for processing test parameters and included files. The logic is in new class - `PrepareTestsCommands` - which is a base class for all prepare commands.

## Removing `--project-dir` argument from Appium and Espresso prepare commands.
Handling this argument is not implemented yet. I'm removing it, so users are not tempted to use it.

### DIFF
--- a/src/commands/test/lib/appium-preparer.ts
+++ b/src/commands/test/lib/appium-preparer.ts
@@ -8,8 +8,6 @@ import * as pfs from "../../../util/misc/promisfied-fs";
 export class AppiumPreparer {
   private readonly artifactsDir: string;
   private buildDir: string;
-  public include: IFileDescriptionJson[];
-  public testParameters: { [key:string]: any };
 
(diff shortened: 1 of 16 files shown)


krukow:
Perfect

PR #69, opened 2016-12-01
Test Cloud: support for relative paths in --include argument

kmiecikt:
In the current version, values of the `--include` argument had to use format `targetPath=sourcePath`.
This change allows to use only single path, as long as it is inside "root directory", i.e.:
- `--build-dir` for Xamarin UI Tests, Espresso and Appium
- `--project-dir` for Calabash

This change also improves validation and unit tests for included files parser.

### DIFF
--- a/src/commands/test/lib/included-files-parser.ts
+++ b/src/commands/test/lib/included-files-parser.ts
@@ -1,17 +1,68 @@
 import { IFileDescriptionJson } from "./test-manifest-reader";
+import * as path from "path";
 
-export function parseIncludedFiles(includedFiles: string[]): IFileDescriptionJson[] {
-  return includedFiles.map(f => parseIncludedFile(f));
+const invalidCharactersRegexp = /['"!#$%&+^<=>`|]/
+
+export function parseIncludedFiles(includedFiles: string[], rootDir: string): IFileDescriptionJson[] {
+  return includedFiles.map(f => parseIncludedFile(f, rootDir));
 }
 
-function parseIncludedFile(includedFile: string): IFileDescriptionJson {
+function parseIncludedFile(includedFile: string, rootDir: string): IFileDescriptionJson {
   let separatorIndex = includedFile.indexOf("=");
   if (separatorIndex === -1) {
-    throw new Error(`Invalid definition of included files: ${includedFile}. Included files must specified in format "targetPath"="sourcePath"`);
+    return parseIncludedFileFromSinglePath(includedFile, rootDir);
+  }
+  else {
+    return parseIncludedFileFromPathsPair(includedFile, rootDir, separatorIndex);
+  }
+}
+
+function parseIncludedFileFromSinglePath(includedFile: string, rootDir: string): IFileDescriptionJson {
+  validatePath(includedFile);
+
+  if (path.isAbsolute(includedFile)) {
+    let targetPath = path.relative(rootDir, includedFile);
+    if (targetPath.indexOf("..") != -1) {
+      throw new Error(`Invalid included file: "${includedFile}". ` + 
+                      `If only a single path is used, it must be inside directory "${rootDir}"`);
+    }
+
+    return {
+      targetPath: path.relative(rootDir, includedFile),
+      sourcePath: includedFile
+    }
+  }
+  else {
+    return {
+      targetPath: includedFile,
+      sourcePath: path.join(rootDir, includedFile)
+    }
+  }
+}
+
+function parseIncludedFileFromPathsPair(includedFile: string, rootDir: string, separatorIndex: number): IFileDescriptionJson {
+  let targetPath = includedFile.substr(0, separatorIndex);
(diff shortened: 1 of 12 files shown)


christav:
@kmiecikt I've merged your other open PRs because they were fairly simple. This one is a little more complex. I took a  look and I think it's fine, but I'd like to have somebody on your team approve as well before merging just because I may have missed something that someone more familiar with testcloud would catch.

test/commands/test/lib/included-files-parser-test.ts
```diff
@@ -1,22 +1,87 @@
...
+  function normalizeFileDescriptions(descriptions: IFileDescriptionJson[]): IFileDescriptionJson[] {
+    return descriptions.map(d => {
+      return {
+        "targetPath": normalizePath(d.targetPath),
+        "sourcePath": normalizePath(d.sourcePath)
+      };
+    })
+  }
+
+  it("should parse pairs with relative target path and valid source path", () => {
```
skovsboll: This one fails on Mac:

```
  1 failing

  1) parseIncludedFiles should parse pairs with relative target path and valid source path:

      AssertionError: expected [ Array(2) ] to deeply equal [ Array(2) ]
      + expected - actual

       [
         {
      -    "sourcePath": "d:/workspace/d:/Temp/Data"
      +    "sourcePath": "d:/Temp/Data"
           "targetPath": "data/foo"
         }
         {
           "sourcePath": "d:/workspace/bar"

      at assertEqual (node_modules/chai/lib/chai/core/assertions.js:485:19)
      at ctx.(anonymous function) (node_modules/chai/lib/chai/utils/addMethod.js:41:25)
      at doAsserterAsyncAndAddThen (node_modules/chai-as-promised/lib/chai-as-promised.js:293:29)
      at .<anonymous> (node_modules/chai-as-promised/lib/chai-as-promised.js:252:17)
      at ctx.(anonymous function) [as equal] (node_modules/chai/lib/chai/utils/overwriteMethod.js:49:33)
      at Context.it (test/commands/test/lib/included-files-parser-test.ts:37:41)
```

test/commands/test/lib/included-files-parser-test.ts
```diff
@@ -1,22 +1,87 @@
...
+      parseIncludedFiles(["/tmp/source"], unitRootDir)
+    }).to.throw();
+  });
+
+  it("should reject string that is not a path", () => {
+    expect(() => {
+      parseIncludedFiles(["|a"], windowsRootDir)
+    }).to.throw();
+  });
 });
```
skovsboll: Here's another edge case you could consider adding:

```
  it("should accept single absolute, deep path under root dir", () => {
    let rawIncludedFiles = [ "/home/user/workspace/somewhere/myAbsoluteData" ];
    let parsedIncludedFiles = normalizeFileDescriptions(parseIncludedFiles(rawIncludedFiles, unitRootDir));

    let expected = [
      {
        "targetPath": "somewhere/myAbsoluteData",
        "sourcePath": "/home/user/workspace/somewhere/myAbsoluteData"
      }
    ];

    expect(parsedIncludedFiles).to.deep.equal(expected);
  });
```

src/commands/test/lib/included-files-parser.ts
```diff
@@ -1,17 +1,68 @@
...
+    return parseIncludedFileFromPathsPair(includedFile, rootDir, separatorIndex);
+  }
+}
+
+function parseIncludedFileFromSinglePath(includedFile: string, rootDir: string): IFileDescriptionJson {
+  validatePath(includedFile);
+
+  if (path.isAbsolute(includedFile)) {
+    let targetPath = path.relative(rootDir, includedFile);
+    if (targetPath.indexOf("..") != -1) {
```
skovsboll: I don't think you have a test for this case -- trying to escape using `..` ?

src/commands/test/lib/included-files-parser.ts
```diff
@@ -1,17 +1,68 @@
...
+  else {
+    return {
+      targetPath: includedFile,
+      sourcePath: path.join(rootDir, includedFile)
+    }
+  }
+}
+
+function parseIncludedFileFromPathsPair(includedFile: string, rootDir: string, separatorIndex: number): IFileDescriptionJson {
+  let targetPath = includedFile.substr(0, separatorIndex);
```
skovsboll: Most of the time, using a package for abstracting these path manipulations can be a good idea. For NodeJS, there's https://nodejs.org/docs/latest/api/path.html. Just a suggestion.

kmiecikt: Yes, I'm already using it. The substr is for parsing "target=source" into two separate paths. Once I have them, I'm using the path module.

test/commands/test/lib/included-files-parser-test.ts
kmiecikt: Fixed

src/commands/test/lib/included-files-parser.ts
kmiecikt: Thanks - added this test

PR #70, opened 2016-12-01
Adding CLI arguments naming convention to contributing.md

kmiecikt:


lumaxis:
@christav @kmiecikt As a further thought on naming conventions: Where applicable, I think CLI parameter names should match existing API parameter names to increase familiarity between the two and make switching easier. I stumbled into this when looking into the app creation command.
Should I add a paragraph for that to the Readme as well?

christav:
@lumaxis How often will the CLI users also be familiar with the Bifrost API? I suspect it'll be "not often." Also, how consistent is naming in our APIs for similar concepts across beacons? I haven't looked but I think it's also "not very".

We can argue over the specific conventions, but I think having conventions like "directory parameter look like this" and "filename parameters look like that" is an excellent one, and helps overall learning and understandability.

In the absence of anything else, sure, aligning with the API sounds great. But if everywhere else in the CLI we've got a convention we should stay consistent with the other CLI commands even if they differentiate from the API naming. Does that make sense?

lumaxis:
Regarding you suspicion regarding people being familiar with both, I'm honestly not sure. When it comes to automation, using the CLI and the API are two very similar tools in your toolbelt and a user might be using one of them locally to automate things and the other on your CI server, for example.

Absolutely, makes perfect sense. I was just thinking that we are currently still at a pretty early stage with the CLI, so everything is pretty new and we are just coming up with naming conventions.

My idea was that if we would want to re-adjust names of CLI commands, a better opportunity for that would be now rather than later.

I agree though that overall, having consistency between different CLI commands trumps single commands being consistent with their API. Maybe we can have both 🙂

PR #73, opened 2016-12-01
Test Cloud: better commands and arguments help messages

kmiecikt:
The biggest change in this PR is introducing new module "help-messages", which stores messages for all Test Cloud commands and arguments. This fixes two problems:
1. Some commands shared arguments names, but used separate help strings which could easy get out of sync.
2. It was very difficult to review help messages, which were scattered around the entire project.

### DIFF
--- a/src/commands/test/lib/help-messages.ts
+++ b/src/commands/test/lib/help-messages.ts
@@ -1,0 +1,32 @@
+import * as os from "os";
+
+export module Messages {
+  export module TestCloud {
+    export module Commands {
+      export const PrepareAppium = "Creates artifacts directory with Appium tests.";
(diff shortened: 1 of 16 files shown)


src/commands/test/lib/help-messages.ts
```diff
@@ -0,0 +1,67 @@
+import * as os from "os";
+
+export module Messages {
+  export module TestCloud {
+    export module Commands {
+      export const PrepareAppium = "Creates artifacts directory with Appium tests.";
+      export const PrepareCalabash = `Creates artifacts directory with Calabash tests.${os.EOL}` +
```
christav: Did you know that backtick-strings can be multiline? You could skip the `${os.EOL}` and explicit concatenation and just use a multiline string. No biggie, but you might want to consider for the future.

PR #74, opened 2016-12-01
Fix broken link to contributing.md in Readme

lumaxis:


christav:
Thanks for finding this, but this is the incorrect fix. See https://help.github.com/articles/relative-links-in-readmes/ for details. The problem with this is that it always points to master, regardless of which branch we're on.

christav:
I went ahead and fixed this in master. Closing this PR.

PR #79, opened 2016-12-07
Add commands for app management

lumaxis:
Opening this PR early to get some feedback, the implementation is not completely done yet though.

Todo items that are on my mind:
- [x] Add debug logs
- [x] Check if all strings are okay and final
- [x] Update `reportApp` function so the `owner` portion looks nicer

### DIFF
--- a/src/commands/apps/delete.ts
+++ b/src/commands/apps/delete.ts
@@ -2,7 +2,7 @@
 
 import { AppCommand, CommandArgs, CommandResult, help, success, failure, ErrorCodes } from "../../util/commandline";
 import { out, prompt } from "../../util/interaction";
-import { MobileCenterClient, models, clientCall } from "../../util/apis";
+import { MobileCenterClient, models, clientRequest } from "../../util/apis";
 
 @help("Delete an app")
 export default class AppDeleteCommand extends AppCommand {
@@ -15,10 +15,16 @@ export default class AppDeleteCommand extends AppCommand {
     const confirmation = await prompt.confirm(`Do you really want to delete the app "${app.identifier}"`);
 
     if (confirmation) {
-      const result = await out.progress("Deleting app ...", clientCall<models.AppResponse>(cb => client.account.deleteApp(app.appName, app.ownerName, cb)));
+      const deleteAppResponse = await out.progress("Deleting app ...", clientRequest<models.AppResponse>(cb => client.account.deleteApp(app.appName, app.ownerName, cb)));
 
-      if (result && (result as any).error.code as string === "NotFound") {
-        return failure(ErrorCodes.NotFound, `the app "${app.identifier}" could not be found`);
+      const statusCode = deleteAppResponse.response.statusCode;
+      if (statusCode >= 400) {
+        switch (statusCode) {
+          case 404:
+            return failure(ErrorCodes.NotFound, `the app "${app.identifier}" could not be found`);
+          default:
+            return failure(ErrorCodes.Exception, "Unknown error when deleting the app");
+        }
       }
     } else {
       out.text(`Deletion of "${app.identifier}" canceled`);
--- a/src/commands/apps/delete.ts
+++ b/src/commands/apps/delete.ts
@@ -21,7 +21,7 @@ export default class AppDeleteCommand extends AppCommand {
       const result = await out.progress("Deleting app ...", clientCall<models.AppResponse>(cb => client.account.deleteApp(app.appName, app.ownerName, cb)));
 
       if (result && (result as any).error.code as string === "NotFound") {
-        return failure(ErrorCodes.InvalidParameter, `the app "${app.identifier}" could not be found`);
+        return failure(ErrorCodes.NotFound, `the app "${app.identifier}" could not be found`);
       }
     } else {
       out.text(`Deletion of "${app.identifier}" canceled`);
--- a/src/commands/apps/delete.ts
+++ b/src/commands/apps/delete.ts
@@ -4,9 +4,6 @@ import { AppCommand, CommandArgs, CommandResult, help, success, failure, ErrorCo
 import { out, prompt } from "../../util/interaction";
 import { MobileCenterClient, models, clientCall } from "../../util/apis";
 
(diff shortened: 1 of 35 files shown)


src/commands/apps/delete.ts
```diff
@@ -0,0 +1,32 @@
...
+  }
+
+  async run(client: MobileCenterClient): Promise<CommandResult> {
+    const app = this.app;
+    const confirmation = await prompt.confirm(`Do you really want to delete the app "${app.identifier}"`);
+
+    if (confirmation) {
+      const result = await out.progress("Deleting app ...", clientCall<models.AppResponse>(cb => client.account.deleteApp(app.appName, app.ownerName, cb)));
+
+      if (result && (result as any).error.code as string === "NotFound") {
```
christav: I don't think you need to check if result exists?

src/commands/apps/show.ts
```diff
@@ -0,0 +1,33 @@
...
+  async run(client: MobileCenterClient): Promise<CommandResult> {
+    const app = this.app;
+
+    const appDetails = await out.progress("Getting app details ...", clientCall<models.AppResponse>(cb => client.account.getApp(app.appName, app.ownerName, cb)));
+
+    switch ((appDetails as any).error.code as string) {
+      case "BadRequest":
+        return failure(ErrorCodes.Exception, "the request was rejected for an unknown reason");
+      case "NotFound":
+        return failure(ErrorCodes.NotLoggedIn, `the app "${app.identifier}" could not be found`);
```
christav: I don't think this is the right return code - this is the "user tried to do something that requires them to be logged in but they aren't" code.

src/util/profile/profile.ts
```diff
@@ -25,14 +25,15 @@ export interface UpdatableProfile {
 }
 
 export interface Profile extends UpdatableProfile {
-  readonly accessToken: Promise<TokenValueType>;
+  readonly accessToken: Promise<string>;
```
christav: Don't change this - the type is here so that when the token type changes in the future we don't need to change a ton of code. And it will change - it'll need to add an expiration time eventually, for example.

src/util/profile/profile.ts
```diff
@@ -100,14 +101,15 @@ class ProfileImpl implements Profile {
   }
 }
 
-const validApp = /^([a-zA-Z0-9-_]{3,100})\/([a-zA-Z0-9-_]{3,100})$/;
+const validApp = /^([a-zA-Z0-9-_.]{1,100})\/([a-zA-Z0-9-_.]{1,100})$/;
```
christav: I got this regex from Sven and it matched what was on the server. Has this changed on the server?

src/util/commandline/app-command.ts
```diff
@@ -45,40 +45,8 @@ export class AppCommand extends Command {
 
     return result;
   }
-}
-
-export function getCurrentApp(optValue: string): ResultOrValue<DefaultApp> {
```
christav: I remember needing this function for something somewhere, but don't remember where. Is it really unused now?

src/util/commandline/app-command.ts
```diff
@@ -11,6 +10,7 @@ export class AppCommand extends Command {
     super(args);
   }
 
+  @shortName("a")
```
christav: We'll need to verify this doesn't overlap with anyone else's existing options. Short options are a valuable commodity. That being said, this is a great candidate for one.

src/util/interaction/prompt.ts
```diff
@@ -13,6 +13,20 @@ export function prompt(message: string): Promise<string> {
 }
 
 export namespace prompt {
+  export function confirm(message: string): Promise<boolean> {
```
christav: Whoops - I thought I'd added this already. Thanks for adding it.

src/util/profile/profile.ts
```diff
@@ -83,7 +84,7 @@ class ProfileImpl implements Profile {
     return this;
   }
 
-  setAccessToken(token: TokenValueType): Promise<Profile> {
+  setAccessToken(token: TokenValueType): Promise<void> {
```
christav: IIRC, I returned the profile here because we need the returned profile with the updated token in it. Subtle errors occurred otherwise.

src/commands/apps/delete.ts
lumaxis: I do because the API returns a `204` on success which results in `result` being `null`

src/util/profile/profile.ts
lumaxis: Yes, the requirements for users and apps have changed recently, this reflects the latest validations on the server.

src/util/profile/profile.ts
lumaxis: I'll check again but I ran into compiler issues with TS 2.1 which apparently has better type checking for generics and errored in this file.

src/commands/apps/show.ts
lumaxis: I'll add `NotFound` for these kinds of errors if that's okay with you 👍

src/util/commandline/app-command.ts
lumaxis: Yea, turns out using `AppCommand` as the superclass of `list` totally doesn't make sense so I can't use `this.app` => Adding `getCurrentApp` back

src/util/profile/profile.ts
lumaxis: Hmm, the getter `get accessToken()` in line 54 actually returns `TokenValueType.token` which is of type `string` so they definitely don't match in one way or the other.

src/util/profile/profile.ts
lumaxis: The actual Promise returns `this` so I updated the function definition to match it

src/commands/apps/delete.ts
christav: Gotcha.

src/util/profile/profile.ts
christav: Is this actually an issue for you? If not, I'm going to digging into the token stuff anyway in the next couple of days (to support --token switch) and can fix this up as part of that work.

PR #86, opened 2016-12-10
Implementing --token switch

christav:


src/commands/lib/logout.ts
```diff
@@ -3,11 +3,14 @@ import { Profile, deleteUser } from "../../util/profile";
 import { out } from "../../util/interaction";
 
 export async function logout(client: MobileCenterClient, user: Profile): Promise<void> {
-  try {
-  await out.progress("Logging out current user...",
-    clientCall(async cb => client.account.deleteApiToken(await user.accessTokenId, cb)));
-  } catch (err) {
-    // Noop, it's ok if deletion fails
+  // Only delete token off the server if CLI created it.
```
lumaxis: Typo?

src/commands/lib/logout.ts
```diff
@@ -3,11 +3,14 @@ import { Profile, deleteUser } from "../../util/profile";
...
+  if (!user.tokenSuppliedByUser) {
+    try {
+    await out.progress("Logging out current user...",
+      clientCall(async cb => client.account.deleteApiToken(await user.accessTokenId, cb)));
+    } catch (err) {
+      // Noop, it's ok if deletion fails
+    }
```
lumaxis: Will this still print an error to the terminal or what will happen in case of error?

src/commands/login.ts
```diff
@@ -68,12 +47,39 @@ export default class LoginCommand extends Command {
...
+      return [{
+        name: "userName",
+        message:"Username or email: "
+      }];
+    }
+    return [];
+  }
+
+  private passwordQuestion(): any[] {
+    if (!this.token && !this.password) {
```
lumaxis: Personally, I would prefer to move this logic to line 37 and only concat the password question if this condition is true. Feels a bit more natural, imo.

src/util/commandline/command.ts
```diff
@@ -101,11 +111,24 @@ export class Command {
...
-    var user = getUser();
-    if (user) {
-      debug(`running commmand logic`);
-      return this.run(createMobileCenterClient(user));
+    if (this.environmentName && !this.token) {
+      return Promise.resolve(Result.illegal("Cannot specify environment without giving token"));
+    }
+
+    let client: MobileCenterClient;
+    if(this.token) {
```
lumaxis: Missing space

PR #89, opened 2016-12-13
Add commands for API token management

lumaxis:


src/commands/tokens/create.ts
```diff
@@ -0,0 +1,49 @@
...
+const debug = require("debug")("mobile-center-cli:commands:apps:create");
+import { inspect } from "util";
+
+@help("Create a new API token")
+export default class TokenCreateCommand extends Command {
+  constructor(args: CommandArgs) {
+    super(args);
+  }
+
+  @help("Description of the app")
```
christav: Shouldn't this be "description of the token"? You're not dealing with an app here.

src/commands/tokens/delete.ts
```diff
@@ -0,0 +1,36 @@
+// tokens delete command
+
+import { Command, CommandArgs, CommandResult, help, success, failure, ErrorCodes, shortName, longName, hasArg, required } from "../../util/commandline";
+import { out, prompt } from "../../util/interaction";
+import { MobileCenterClient, models, clientRequest } from "../../util/apis";
+
+const debug = require("debug")("mobile-center-cli:commands:apps:create");
+import { inspect } from "util";
+
+@help("Delete an app")
```
christav: Again, delete the token, not app.

src/commands/tokens/list.ts
```diff
@@ -0,0 +1,24 @@
...
+  constructor(args: CommandArgs) {
+    super(args);
+  }
+
+
+  async run(client: MobileCenterClient): Promise<CommandResult> {
+    const apiTokensResponse = await out.progress("Getting API tokens ...",
+      clientRequest<models.ApiTokenResponse[]>(cb => client.account.getApiTokens(cb)));
+
+    apiTokensResponse.result.map(apiToken => reportTokenInfo(apiToken));
```
christav: For future consideration - use out.table to display this instead of out.report. Or use out.report but pass the entire array in one shot instead of the external map call.

src/util/interaction/out.ts
```diff
@@ -36,6 +36,8 @@ export function progress<T>(title: string, action: Promise<T>): Promise<T> {
 // function.
 //
 export function list<T>(formatter: {(item: T): string}, items: T[]): void {
+  if (!items || Object.keys(items).length === 0) { return; }
```
christav: This check seems a little weird for a function that's supposed to take an array - why are you checking `Object.keys(items).length`?

swagger/bifrost.swagger.json
```diff
@@ -2,10 +2,10 @@
   "swagger": "2.0",
```
christav: If you're updating this file, please check in the `.before.` file as well and the generated client code as well.

src/util/interaction/out.ts
lumaxis: I had an issue where the response I got didn't match the responses know to the generate API client which resulted in a `{}` being returned. `Object.keys(items).length` should catch both cases, `Object` and `Array`.

swagger/bifrost.swagger.json
lumaxis: Ah, of course. Added!

src/commands/tokens/list.ts
lumaxis: `out.table` looks much nicer 👌

PR #96, opened 2017-01-10
Test Cloud: faster upload protocol

kmiecikt:
This change improves upload speed for file hashes. By sending all file hashes in one request, it can reduce upload speed for Appium tests from 27 to 16 seconds (and put much less pressure on Bifrost).


gulpfile.js
```diff
@@ -69,7 +69,7 @@ gulp.task('fixup-swagger', function () {
   autorest.fixupRawSwagger('./swagger/bifrost.swagger.before.json', './swagger/bifrost.swagger.json');
 });
 
-gulp.task('autorest', ['clean-autorest', 'fixup-swagger'], function () {
```
christav: Without the clean-autorest target in here, we could get leftover files in the client code if for example operations were removed from the swagger. Why did you take this out?

src/commands/test/lib/test-cloud-uploader.ts
```diff
@@ -123,52 +124,44 @@ export class TestCloudUploader {
...
-    }
-    else {
-      await this.uploadFile(testRunId, file);
-      debug(`File ${file.sourcePath}: direct upload`);
-    }
+  private async uploadFilesUsingBatch(testRunId: string, files: TestRunFile[]): Promise<void> {
+    let checkHashResult = await this.uploadHashesBatch(testRunId, files.map(f => { return { file: f }; }));
+
+    let limit = pLimit(paralleRequests);
+    let uploadNewFilesTask = checkHashResult
```
christav: If this is an array, it should probably have a plural name - `uploadNewFilesTasks` instead?

src/commands/test/lib/test-cloud-uploader.ts
```diff
@@ -123,52 +124,44 @@ export class TestCloudUploader {
...
+      .map(r => limit(() => this.uploadFile(testRunId, r.file)));
+
+    await Promise.all(uploadNewFilesTask);
   }
 
-  private async tryUploadFileHash(testRunId: string, file: TestRunFile, byteRange: string = null): Promise<boolean> {
-    let response = await new Promise<http.IncomingMessage>((resolve, reject) => {
-      this._client.test.uploadHash(
+  private async uploadHashesBatch(testRunId: string, files: { file: TestRunFile, byteRange?: string }[]): Promise<{ file: TestRunFile, response: models.TestCloudFileHashResponse }[]> { 
+    let hashResponses = await new Promise<models.TestCloudFileHashResponse[]>((resolve, reject) => {
```
christav: We have new new `clientRequest` wrapper that gives you access to the actual HTTP response as well as the result, can you use that now instead of manual promise wrangling?

gulpfile.js
kmiecikt: I have reverted that, but I'm confused: how do I generate new api from updated swagger file?
Some time ago, I was updating bifrost.swagger.json, and the calling gulp autorest. If I do it right now, my changes in bifrost.swagger.json are reverted.

src/commands/test/lib/test-cloud-uploader.ts
kmiecikt: Done

gulpfile.js
christav: Ah, sure. That behavior is because of the `fixup-swagger` task. Basically, I found the changes we needed to make (so far anyway) to the bifrost swagger document could be automated, so that's what it does.

What sort of changes are you making? If they're just the default required ones (adding operation IDs, removing empty paths, one other thing I don't remember ;) ) you can drop the swagger from bifrost in at the *.before.* file name and everything will just work.

If you need to do more/different than that ping me on slack/email and we'll work out the process.

Sorry about the confusion, I should have been louder about this change.

src/commands/test/lib/test-cloud-uploader.ts
kmiecikt: Done

PR #97, opened 2017-01-17
Fixing error message for CalabashPreparer

kmiecikt:


christav:
One thing to consider for these error messages - they say "please see logs for details". Which logs are they talking about? Some info about where to find the logs (or how to turn them on) would be great.

PR #99, opened 2017-02-02
Fixed annotation on id property for tokens delete command.

christav:
Command failed with "missing required parameter" without it.


lumaxis:
Whoops, sorry about that! Fairly certain I tested this though, strange 🤔

christav:
I was checking, discovered I had like 30+ old api tokens defined, and tried to delete a couple. And it failed.
No big, and I just released 0.4.0 so it's fixed in the file now.

Issue #93, opened 2016-12-14
Default explanation text for `mobile-center test` is poorly formatted


dvdsgl:
`mobile-center test` produces this output:

![image](https://cloud.githubusercontent.com/assets/108197/21202605/e4147534-c203-11e6-86bc-24052db1e603.png)

christav:
I've done some investigation. The help system is properly pulling the information to display out, but the `cli-table2` module that's being used for formatting isn't building the table correctly. I'm guessing there's issues when the width of the table is larger than the width of the terminal.

I was thinking of trying to set an explicit size on the table columns instead of letting it pick them automatically and see if that fixes the problem. If it doesn't we'll need to write a bunch of manual formatting stuff.

christav:
Fixed in #101

PR #101, opened 2017-02-03
Default explanation text for `mobile-center test` is poorly formatted

bryukov:
Fix for #93 

1. Fixed the explanation text formatting by setting the `colWidths` on the `noTableBorders` settings set. `colWidths` are calculated from the current console width.

2. Fixed another problem which also broke output formatting in some consoles (e.g. cmder and powershell). It was caused by the presence of the CRLF characters in some of the category.txt files. cli-table2 doesn't take CR into account (i.e. doesn't removes it when wrapping strings to fit into table cell), thus breaking the formatting.

Here is how it looks like after the fix (powershell console as an example):

powershell

Please, note that there is no space between rightmost top cell and the one below it (see the red line on the next image):

powershell_redline

Is it by design? Or should I add a horizontal space between them?

### DIFF
--- a/src/util/interaction/out.ts
+++ b/src/util/interaction/out.ts
@@ -10,6 +10,8 @@ import * as wrap from "wordwrap";
 const Table = require("cli-table2");
 const Spinner = require("cli-spinner").Spinner;
 
+import { terminal } from "./terminal";
+
 //
 // Display a progress spinner while waiting for the provided promise
 // to complete.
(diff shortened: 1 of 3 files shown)


christav:
Please add a space between the two categories - it's too hard to read without it.

src/util/interaction/out.ts
```diff
@@ -116,14 +118,26 @@ export function table(...args: any[]): void {
...
-  },
-  style: { "padding-left": 0, "padding-right": 0 }
+export function getNoTableBordersSetting(tableColumnsCount: number) {
+  let consoleWidth = terminal.columns();
+  // There is a single whitespace to the right from the each column, count it as unavailable
+  let availableWidth = consoleWidth - tableColumnsCount;
+  let colWidth = Math.floor(availableWidth / tableColumnsCount);
+  let remainderWidth = availableWidth % tableColumnsCount;
+  // Spreading remainder width across the first columns
+  let colWidths = Array<number>(remainderWidth).fill(colWidth + 1).concat(Array<number>(tableColumnsCount - remainderWidth).fill(colWidth));
```
christav: I don't think we want to evenly divide the space - the left column in this case (the category name) is probably narrower than the text. Perhaps pass in the desired width for the first column and let the rest of the space go to the right column?

christav: Also, I'm fine leaving this function hard-wired for the help system at the moment.

bryukov:
@christav I have addressed your comments in another commit. Can you, please, take a look on it?

The help messages for "test" and "apps set-current" after the changes look like this:
iteration2

PR #108, opened 2017-02-09
Avoid throwing errors in command constructors

jonstoneman:
Otherwise:
- nothing is output
- the help command is broken

### DIFF
--- a/src/commands/test/lib/prepare-tests-command.ts
+++ b/src/commands/test/lib/prepare-tests-command.ts
@@ -41,8 +41,13 @@ export class PrepareTestsCommand extends Command {
     }
   }
 
+  // Override this for things like parameter validation to avoid throwing errors in the constructor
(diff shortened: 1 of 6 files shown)


src/commands/test/lib/prepare-tests-command.ts
```diff
@@ -41,8 +41,13 @@ export class PrepareTestsCommand extends Command {
     }
   }
 
+  // Override this for things like parameter validation to avoid throwing errors in the constructor
+  protected async runNoClientInner(): Promise<void> {
```
christav: Wouldn't this be better named `validateOptions` or something along those lines?

jonstoneman: I went with a generic name as the method could be used for other things in the future.  I guess it's better having a more specific name and either change it or add other methods if needed.

PR #124, opened 2017-02-28
Make (only) required parameters required

jonstoneman:


src/commands/test/prepare/espresso.ts
```diff
@@ -10,6 +10,7 @@ export default class PrepareEspressoCommand extends PrepareTestsCommand {
   @help(Messages.TestCloud.Arguments.EspressoBuildDir)
   @longName("build-dir")
   @hasArg
+  @required
```
kmiecikt: You can use either `build-dir` or `test-apk-path`, so it should not be required.

jonstoneman: good catch - thanks

PR #125, opened 2017-02-28
Refactor `--app-path` to simplify code

jonstoneman:
No functional changes but I prefer this approach


src/commands/test/lib/run-tests-command.ts
```diff
@@ -82,6 +89,10 @@ export abstract class RunTestsCommand extends AppCommand {
   }
 
   public async run(client: MobileCenterClient): Promise<CommandResult> {
+    if (this.isAppPathRquired && !this.appPath)
+    {
```
kmiecikt: Nitpick: rest of the code puts `{` at the same line as the `if` statement.

PR #127, opened 2017-03-02
CalabashPrepare: fixing regression when no test parameters are provided

kmiecikt:
Fixes a regression introduced with a recent fix for test parameters.

With the latest fix, we pass empty test parameters array to the `CalabashPreparer` if no parameters are used. However, the preparer was not handling it correctly, and generated empty `--test-parameters` argument passed to the Ruby CLI `test-cloud`. The Ruby CLI couldn't parse it, and returned a failure.

### DIFF
--- a/src/commands/test/prepare/calabash.ts
+++ b/src/commands/test/prepare/calabash.ts
@@ -66,6 +66,10 @@ export default class PrepareCalabashCommand extends PrepareTestsCommand {
       this.configPath = this.config;
     }
 
+    if (this.config && this.configPath) {
+      throw new Error("Both arguments --config-path and --config (obsolete) were used. Please use only --config-path.")
+    }
+
     if (!this.projectDir) {
       throw new Error("Argument --project-dir is required");
     }
--- a/src/commands/test/prepare/calabash.ts
+++ b/src/commands/test/prepare/calabash.ts
@@ -32,10 +32,15 @@ export default class PrepareCalabashCommand extends PrepareTestsCommand {
   @hasArg
   signInfo: string;
 
-  @help(Messages.TestCloud.Arguments.CalabashConfigPath)
+  @help("Obsolete. Please use --config-path instead")
   @longName("config")
   @hasArg
   config: string;
+
+  @help(Messages.TestCloud.Arguments.CalabashConfigPath)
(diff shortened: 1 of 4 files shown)


jonstoneman:
LGTM

src/commands/test/prepare/calabash.ts
```diff
@@ -56,6 +61,15 @@ export default class PrepareCalabashCommand extends PrepareTestsCommand {
       this.projectDir = this.workspaceDir;
     }
 
+    if (this.config && this.configPath) {
+      throw new Error("Both arguments --config-path and --config (obsolete) were used. Please use only --config-path.")
```
jonstoneman: I think `Arguments --config-path and --config (obsolete) were both used. Please use only --config-path.` reads better.

src/commands/test/prepare/calabash.ts
```diff
@@ -56,6 +61,15 @@ export default class PrepareCalabashCommand extends PrepareTestsCommand {
...
+    }
+
+    if (this.config && !this.configPath) {
```
jonstoneman: This could be just `if (this.config) {` now that the above check has been added.

src/commands/test/prepare/calabash.ts
kmiecikt: Done

src/commands/test/prepare/calabash.ts
kmiecikt: Done

src/util/misc/process-helper.ts
```diff
@@ -1,9 +1,17 @@
...
 import { out } from "../interaction";
 
-export function execAndWait(command: string): Promise<number> {
+export function execAndWait(command: string, onStdOut?: (text: string) => void, onStdErr?: (text: string) => void): Promise<number> {
   return new Promise((resolve, reject) => {
+    if (!onStdOut) {
+      onStdOut = text => out.text(text);
+    }
+    if (!onStdErr) {
+      onStdErr = text => out.text(text);
```
christav: Should we add an `out.warn` or `out.err` or something like that? With the idea that using it would output to stderr instead of stdout?

PR #128, opened 2017-03-03
Use promisfied-glob instead of glob

jonstoneman:


src/commands/test/lib/espresso-preparer.ts
```diff
@@ -82,7 +69,7 @@ export class EspressoPreparer {
 
   private async detectTestApkPathFromBuildDir(): Promise<string> {
     let apkPattern = path.join(this.buildDir,"*androidTest.apk");
```
kmiecikt: Can you add space here (not related to your changes)?

PR #130, opened 2017-03-03
Implemented "build download" command

bryukov:
Implemented "build download" command.
capture

This PR depends on https://github.com/Microsoft/mobile-center-cli/pull/129

PR #134, opened 2017-03-07
Implemented 'build queue' command

bryukov:
Implemented 'build queue' command (including build portal URL generation):
capture

Depends on #130 for 'build' category help


src/commands/build/queue.ts
```diff
@@ -0,0 +1,54 @@
...
+        debug: this.debugLogs
+      }, cb)));
+
+    const queueBuildHttpResponseCode = queueBuildRequestResponse.response.statusCode;
+
+    if (queueBuildHttpResponseCode >= 400) {
+      return failure(ErrorCodes.Exception, "the Queue Build request was rejected for an unknown reason");
+    }
+
+    // TODO Result is processed incorrectly now, so workaround is needed to get build ID and real branch name.
```
bryukov: Once result processing will be fixed, the following two lines can be uncommented and used instead of the workaround.

PR #136, opened 2017-03-08
Add test run/prepare xcuitest command

jonstoneman:


src/commands/test/lib/help-messages.ts
```diff
@@ -16,7 +16,8 @@ export module Messages {
                                     `Required external tools:${os.EOL}` + 
                                     `- .NET Framework on Windows, Mono Runtime on OS X${os.EOL}` + 
                                     `- NuGet package Xamarin.UITests, version 2.0.1 or higher`;
-
+      export const PrepareXCUITest = "Creates an artifacts directory with XCUITest tests.";
```
kmiecikt: Can you also add _an_ to help messages for other frameworks?

src/commands/test/lib/xcuitest-preparer.ts
```diff
@@ -0,0 +1,93 @@
...
+      throw new Error("Arguments buildDir and testIpaPath cannot be used together");
+    }
+
+    this.artifactsDir = artifactsDir;
+    this.buildDir = buildDir;
+    this.testIpaPath = testIpaPath;
+  }
+
+  public async prepare(): Promise<string> {
+    if (!pfs.exists(this.artifactsDir)) {
```
kmiecikt: Should be `if (!(await pfx.exists(...)))`

kmiecikt: C# compiler is much better at detecting tasks that are not awaited. Also, the type checking would prevent you from converting Task / Promise to Boolean.

src/commands/test/lib/xcuitest-preparer.ts
```diff
@@ -0,0 +1,93 @@
...
+  
+    let manifestPath = path.join(this.artifactsDir, "manifest.json");
+    let manifest = await this.createXCUITestManifest();
+    let manifestJson = JSON.stringify(manifest, null, 1);
+    await pfs.writeFile(manifestPath, manifestJson);
+
+    return manifestPath;
+  }
+
+  private async validatePathExists(path: string, isFile: boolean, errorMessage: string): Promise<void> {
```
kmiecikt: What about moving it to `promisfied-fs.ts` and renaming to something like `fileExists`?

src/commands/test/lib/xcuitest-preparer.ts
jonstoneman: good catch!

jonstoneman:
@kmiecikt: How's that look now?

skovsboll:
@kmiecikt could you review, please?

src/commands/test/lib/help-messages.ts
```diff
@@ -40,35 +42,38 @@ export module Messages {
...
-      export const UITestsToolsDir = "Path to directory with Xamarin UI Tests tools that contains test-cloud.exe";
+      export const UITestsToolsDir = "Path to the directory containing the Xamarin UI Tests tools including test-cloud.exe";
+
+      export const StatusTestRunId = "ID of the started test run";
+      export const StatusContinuous = "Continuously checks the test run status until it completes";
 
-      export const StatusTestRunId = "ID of started test run";
-      export const StatusContinuous = "Continuously checks test run status until it completes";
+      export const XCUITestIpaPath = "Path to the *.ipa file with the XCUITest tests";
+      export const XCUITestBuildDir = "Path to the XCUITest output directory (usually <project>/Build/Products/Debug-iphoneos)";
```
sapieneptus: I assume this is a user-facing string, no? I'm not sure 'XCUITest output directory' is totally accurate (did I use that phrase somewhere? Apologies if I did). 

The more accurate phrase would be 'the `build-for-testing` output directory' but I'm not sure if that sounds confusing? It refers to the `derivedDataPath` directory when running `xcodebuild build-for-testing`

src/commands/test/lib/xcuitest-preparer.ts
```diff
@@ -0,0 +1,78 @@
...
+  constructor(artifactsDir: string, buildDir: string, testIpaPath: string) {
+    if (!artifactsDir) {
+      throw new Error("Argument artifactsDir is required");
+    }
+
+    if (!(buildDir || testIpaPath)) {
+      throw new Error("Either buildDir or testIpaPath argument is required");
+    }
+
+    if (buildDir && testIpaPath) {
```
sapieneptus: I thought we were supporting this format?

xtc uploader:
```
xtc xcuitest --app-path ./aut.ipa <args> --workspace /Build/Products/Debug-iphoneos
```

The use-case is if someone is writing tests for an app for which they don't have the source code.

src/commands/test/lib/xcuitest-preparer.ts
```diff
@@ -0,0 +1,78 @@
...
+  private async generateTestIpa(): Promise<void> {
+    let runnerAppPaths = await pglob.glob(path.join(this.buildDir, "*-Runner.app"));
+    if (runnerAppPaths.length == 0) {
+      throw new TestCloudError(`Unable to find test runner app within ${this.buildDir}`);
+    }
+    if (runnerAppPaths.length > 1) {
+      throw new TestCloudError(`Multiple test runner apps found within ${this.buildDir}`);
+    }
+    this.testIpaPath = path.join(this.artifactsDir, `${path.parse(runnerAppPaths[0]).name}.ipa`);
+    await iba.archiveAppBundle(runnerAppPaths[0], this.testIpaPath);
```
sapieneptus: stealing my method names!

src/commands/test/lib/xcuitest-preparer.ts
sapieneptus: Oh wait, `testIpaPath` refers to the runner?

I've been trying to be strict about referring to the two apps as `runner` and `aut`, because 'testApp' I find quite ambiguous in this case.

src/commands/test/run/xcuitest.ts
```diff
@@ -0,0 +1,74 @@
...
+    }
+  }
+
+  protected getSourceRootDir(): string {
+    return this.buildDir;
+  }
+
+  private async generateAppIpa(): Promise<void> {
+    let appPaths = (await pfs.readdir(this.buildDir)).filter((appPath) => {
+      if (appPath.endsWith("-Runner.app")) {
```
sapieneptus: Oh man, there's no way to just regex this?
```
/^[^-Runner].app$/
```

sapieneptus:
Looks good as far as I can tell (i.e. matches specs from the testcloud-uploader), but I would request some consideration of changing the copy/variable names as mentioned.

src/commands/test/lib/help-messages.ts
jonstoneman: How about just "the build output" directory?  I think "usually <project>/Build/Products/Debug-iphoneos" gives enough of a clue as to what that is..

sapieneptus: I think that would work 👍

src/commands/test/lib/xcuitest-preparer.ts
jonstoneman: In mobile-center, the AUT is always `appPath` so I think the confusion is somewhat less in this case.  `testIpaPath` matches the command line parameter `test-ipa-path` and, while `test-runner-ipa-path` would be clearer, it's also a bit ugly and inconvenient for users.

@kmiecikt What do you think?

sapieneptus: I think the important thing is to have consistent standards either way. Currently in the pipeline job we've followed the `runner/aut` convention. @john7doe any opinions?

src/commands/test/run/xcuitest.ts
jonstoneman: I can't find a nicer way of doing this.  I could `glob(*.app)` and then run the results through a regex but I prefer what I have.

@kmiecikt do you know of any better option?

src/commands/test/lib/help-messages.ts
jonstoneman: 👍  Oh, and I don't think you did use the term `XCUITest output directory` btw..

src/commands/test/run/xcuitest.ts
kmiecikt: What about something like `glob(``${this.buildDir}/*-Runner.app``)`?

src/commands/test/lib/xcuitest-preparer.ts
kmiecikt: I think it should be `test-ipa-path`. That will be consistent with `test-apk-path` that we have in `run espresso` and `prepare espresso` commands.

src/commands/test/run/xcuitest.ts
jonstoneman: We actually want to match `${this.buildDir}/*.app` but not `${this.buildDir}/*-Runner.app`.  

I couldn't get it to work using `glob`, although something like the following _should_ have worked (for some versions of glob) according to the docs I found:
- ``glob(`${this.buildDir}/*!(-Runner).app`)``
- ``glob(`${this.buildDir}/*.app`, {"ignore": [ "*-Runner.app" ] })``

jonstoneman:
This is ready to merge subject to review by @christav and if anyone can suggest a better way of file matching: https://github.com/Microsoft/mobile-center-cli/pull/136#discussion_r105947550

src/commands/test/lib/xcuitest-preparer.ts
sapieneptus: Well, I'll just go on the record saying I think it's ambiguous. We have have (many) cases of people uploading the AUT in place of the runner, and the term `test-app/ipa/apk` seems inherently ambiguous when one app is the "tester" and the other is the "tested". 

For me, personally, I wouldn't know which `test-ipa-path` should refer to without some additional description. Maybe I'm alone in that opinion, but just thought I'd express it.

sapieneptus:
@jonstoneman Just sanity checking, but the scope of this PR is just to upload the XCUITest runner + AUT and rely on the XTC default .xctestconfiguration, correct?

We will at some point (like today) be adding a flag for specifying a custom .xctestconfiguration file to the 'legacy' test-cloud-uploader. @krukow Is that functionality considered V1.0 essential? Just trying to make sure we don't overlook it if it needs to be in this PR as well.

src/commands/test/run/xcuitest.ts
sapieneptus: @jonstoneman @kmiecikt 

I know _nothing_ about `typescript`, but will this not work?
```typescript
let appPaths = (await pfs.readdir(this.buildDir)).filter((appPath) => {
    regexp = new RegExp('^[^-Runner].app$'),
    return regexp.test(appPath);
});
```

sapieneptus: In fact, [I can prove it works](http://www.typescriptlang.org/play/#src=%0D%0Alet%20apps%20%3D%20%5B'test.app'%2C%20'test-Runner.app'%5D%3B%0D%0A%0D%0Alet%20appPaths%20%3D%20(apps.filter((appPath)%20%3D%3E%20%7B%0D%0A%20%20%20%20let%20regexp%20%3D%20new%20RegExp(%22%5B%5E-Runner%5D.app%22)%3B%0D%0A%20%20%20%20return%20regexp.test(appPath)%3B%0D%0A%7D))%3B%0D%0A%0D%0Aalert(%22App%20paths%20%3D%20%22%20%2B%20appPaths)%3B);

jonstoneman:
@sapieneptus: I was expecting that to be a future feature.  The plan was to wait for it to be added to test-cloud-uploader before copying the functionality in the cli.

If needed, I'm sure I could add it to the cli before it's added to test-cloud-uploader but I'll need full details - I don't even know what a `.xctestconfiguration` is at the moment.

src/commands/test/run/xcuitest.ts
jonstoneman: Thanks @sapieneptus - not sure why I didn't consider using the regex with the `filter()`!

Also found you can do regexs like this...

```Typescript
let appPaths = (await pfs.readdir(this.buildDir)).filter(
  (appPath) => /[^-Runner].app$/.test(appPath)
);
```

👍

krukow:
@jonstoneman @christav should we get this merged to prepare for release?

src/util/misc/promisfied-fs.ts
```diff
@@ -129,6 +129,27 @@ export function openTempFile(affixes: string): Promise<{path: string, fd: number
...
+  let stats: fs.Stats = null;
+  
+  try {
+    stats = await stat(path);
+  }
+  catch (err) {
+    return false;
+  }
+
+  return isFile === stats.isFile();
```
christav: This might break if they point to a symlink rather than a real file.

PR #135, opened 2017-03-07
Implemented 'build branches show' command

bryukov:
Implemented 'build branches show' command (including adding 'build branches' category help):
capture

Depends on #134 for the portal URL generation

PR #137, opened 2017-03-09
Implemented 'build branches list' command

bryukov:
Implemented 'build branches list' command:
capture

Refactored 'build branches show' to extract common output logic.

Depends on #135 for 'build branches show' implementation

### DIFF
--- a/src/util/apis/generated/operations/buildOperations.js
+++ b/src/util/apis/generated/operations/buildOperations.js
@@ -860,7 +860,7 @@ BuildOperations.prototype.getCommits = function (shaCollection, ownerName, appNa
   // Construct URL
   var baseUrl = this.client.baseUri;
   var requestUrl = baseUrl + (baseUrl.endsWith('/') ? '' : '/') + 'v0.1/apps/{owner_name}/{app_name}/commits/batch/{sha_collection}';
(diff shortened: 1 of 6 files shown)


src/util/apis/generated/operations/buildOperations.js
```diff
@@ -860,7 +860,7 @@ BuildOperations.prototype.getCommits = function (shaCollection, ownerName, appNa
   // Construct URL
   var baseUrl = this.client.baseUri;
   var requestUrl = baseUrl + (baseUrl.endsWith('/') ? '' : '/') + 'v0.1/apps/{owner_name}/{app_name}/commits/batch/{sha_collection}';
-  requestUrl = requestUrl.replace('{sha_collection}', encodeURIComponent(shaCollection));
```
bryukov: This is not a real fix, it is merely a temporary workaround. We need to remove it before pushing this PR. Ideally the rest of the changes from this commit will not require any updates, once buildOperations.getCommits will be fixed for a multiple SHAs.

christav: I took a look, there's a switch in autorest we can use to turn off the url encoding instead of having to manually patch the code. I'll add a thing to the swagger fixup script to add this manually while I yell at build to update their swagger.

PR #138, opened 2017-03-14
Use node fetch instead of promisfied-https

jonstoneman:
@christav: This probably shouldn't be merged as-is as I had to change `import * as fetch from 'node-fetch'` to `const fetch = require('node-fetch')` to avoid this error:
```
error TS2349: Cannot invoke an expression whose type lacks a call signature. Type 'typeof 'node-fetch'' has no compatible call signatures. 
```


jonstoneman:
@christav  👍    Do you have any advice to help me tackle the typings issues or is that something you should look at?

lumaxis:
@jonstoneman The issue was because `fetch` is using a [default export](https://www.typescriptlang.org/docs/handbook/modules.html#default-exports).
`import fetch from 'node-fetch';` works.

jonstoneman:
Thanks Lukas!  

@christav I've made that change and everything (inc intelisense) is working so I guess this should be good to merge now..

christav:
Going to close this for now, will revisit later.

PR #141, opened 2017-03-15
Implemented 'distribute release' command

bryukov:
Implemented command according to the specifications.
capture

### DIFF
--- a/src/commands/distribute/release.ts
+++ b/src/commands/distribute/release.ts
@@ -1,0 +1,261 @@
+import { AppCommand, CommandResult, ErrorCodes, failure, hasArg, help, isCommandFailedResult, longName, required, shortName, success} from "../../util/commandline";
+import { MobileCenterClient, models, clientRequest, ClientResponse } from "../../util/apis";
+import { out } from "../../util/interaction";
+import { inspect } from "util";
+import * as _ from "lodash";
+import * as Request from "request";
+import * as Path from "path";
+import * as Pfs from "../../util/misc/promisfied-fs";
+import { DefaultApp } from "../../util/profile";
+
+const debug = require("debug")("mobile-center-cli:commands:distribute:release");
+
+@help("Upload release binary and trigger distribution")
+export default class ReleaseBinaryCommand extends AppCommand {
+  @help("Path to binary file")
+  @shortName("f")
+  @longName("file")
+  @required
+  @hasArg
+  public filePath: string;
+
+  @help("Distribution group name")
+  @shortName("g")
+  @longName("group")
+  @required
+  @hasArg
+  public distributionGroup: string;
+
+  @help("Release notes text")
+  @shortName("r")
+  @longName("release-notes")
+  @hasArg
+  public releaseNotes: string;
+
+  @help("Path to release notes file")
+  @shortName("R")
+  @longName("release-notes-file")
+  @hasArg
+  public releaseNotesFile: string;
+
+  public async run(client: MobileCenterClient): Promise<CommandResult> {
+    try {
+      return await this.doCommand(client);
+    } catch (error) {
+      if (isCommandFailedResult(error)) {
+        return error;
+      } else {
+        throw error;
+      }
+    }
+  }
+
+  private async doCommand(client: MobileCenterClient): Promise<CommandResult> {
+    const app: DefaultApp = this.app;
+    
+    debug("Check that user hasn't selected both --release-notes and --release-notes-file");
+    this.validateParameters();
+
+    debug("Loading prerequisites");
+    const [distributionGroupUsersCount, releaseBinaryFileBuffer, releaseNotesString] = await out.progress("Loading prerequisites...", this.getPrerequisites(client));
+
+    debug("Creating release upload");
+    const createdReleaseUpload = await this.createReleaseUpload(client, app);
+    const uploadUri = createdReleaseUpload.uploadUrl;
+    const uploadId = createdReleaseUpload.uploadId;
+
+    let releaseUrl: string;
+    try {
+      debug("Uploading release binary");
+      await out.progress("Uploading release binary...", this.uploadFileToUri(uploadUri, releaseBinaryFileBuffer, Path.basename(this.filePath)));
+
+      debug("Finishing release upload");
+      releaseUrl = await this.finishReleaseUpload(client, app, uploadId);
+    } catch (error) {
+      try {
+        out.text("Release upload failed");
+        await this.abortReleaseUpload(client, app, uploadId);
+        out.text("Release upload was aborted");
+      } catch (abortError) {
+        debug("Failed to abort release upload");
+      }
+
+      throw error;
+    }
+
+    debug("Extracting release ID from the release URL");
+    const releaseId = this.extractReleaseId(releaseUrl);
+
+    debug("Distributing the release");
+    const releaseDetails = await this.distributeRelease(client, app, releaseId, releaseNotesString);
+
+    if (_.isNull(distributionGroupUsersCount)) {
+      out.text((rd) => `Release ${rd.shortVersion} (${rd.version}) was successfully released to ${this.distributionGroup}`, releaseDetails);
+    } else {
+      out.text((rd) => `Release ${rd.shortVersion} (${rd.version}) was successfully released to ${distributionGroupUsersCount} testers in ${this.distributionGroup}`, releaseDetails);
+    }
+
+    return success();
+  }
+
+  private validateParameters(): void {
+    if (!_.isNil(this.releaseNotes) && !_.isNil(this.releaseNotesFile)) {
+      throw failure(ErrorCodes.InvalidParameter, "'--release-notes' and '--release-notes-file' switches are mutually exclusive");
+    }
+  }
+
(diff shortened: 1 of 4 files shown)


src/commands/distribute/release.ts
```diff
@@ -0,0 +1,224 @@
...
+      throw failure(ErrorCodes.Exception, `failed to create release upload for ${this.filePath}`);
+    } else {
+      return createReleaseUploadRequestResponse.result;
+    }
+  }
+
+  private uploadFileToUri(uploadUrl: string, fileBuffer: Buffer, filename: string): Promise<void> {
+    debug("Uploading the release binary");
+    return new Promise<void>((resolve, reject) => {
+      Request.post({
```
izikl: why not using async await here?

src/commands/distribute/release.ts
```diff
@@ -0,0 +1,224 @@
...
+    } catch (error) {
+      try {
+        out.text("Release upload failed");        
+        await this.abortReleaseUpload(client, app, uploadId);
+        out.text("Release Upload was aborted");
+      } catch (abortError) {
+        out.text("Failed to abort release upload");
+      }
+
+      throw error;
```
izikl: well the error be printed to the user?

src/commands/distribute/release.ts
```diff
@@ -0,0 +1,224 @@
...
+    }
+
+    debug("Extracting release ID from the release URL");
+    const releaseId = this.extractReleaseId(releaseUrl);
```
izikl: 1.0 (45) === short_version (version)

src/commands/distribute/release.ts
```diff
@@ -0,0 +1,224 @@
...
+  private async abortReleaseUpload(client: MobileCenterClient, app: DefaultApp, uploadId: string): Promise<void> {
+    const abortReleaseUploadRequestResponse = await out.progress("Aborting release upload...", 
+      clientRequest<models.ReleaseUploadEndResponse>((cb) => client.distribute.updateReleaseUpload(uploadId, app.ownerName, app.appName, "aborted", cb)));
+    
+    if (abortReleaseUploadRequestResponse.response.statusCode >= 400) {
+      throw new Error(`HTTP ${abortReleaseUploadRequestResponse.response.statusCode} - ${abortReleaseUploadRequestResponse.response.statusMessage}`);
+    }
+  }
+
+  private extractReleaseId(releaseUrl: string): number {
```
izikl: Err.... yach. Now I understand the motivation for you question in the email. 
I didn't know the console methods doesn't allow passing the resource URL. 
I will might end up just adding the release id to the update upload response.

src/commands/distribute/release.ts
```diff
@@ -0,0 +1,224 @@
...
+      throw error;
+    }
+
+    debug("Extracting release ID from the release URL");
+    const releaseId = this.extractReleaseId(releaseUrl);
+
+    debug("Distributing the release");
+    const releaseDetails = await this.distributeRelease(client, app, releaseId, releaseNotesString);    
+
+    out.text(`Release ${releaseDetails.shortVersion} (${releaseDetails.id}) was successfully released to ${distributionGroupUsers.length} testers of ${this.distributionGroup}`);
```
izikl: `in not `of`: testers in ${this.distributionGroup}

src/commands/distribute/release.ts
```diff
@@ -0,0 +1,224 @@
...
+        if (error.code === "ENOENT") {
+          throw failure(ErrorCodes.InvalidParameter, `release notes file '${this.releaseNotesFile}' doesn't exist`);
+        } else {
+          throw error;
+        }
+      })
+     : Promise.resolve(this.releaseNotes);
+    
+    // get distribution group users (and check distribution group existence)
+    const distributionGroupUsers = this.getDistributionGroupUsers(client);
```
izikl: to make it clear that DGs are not used, maybe just return the length here.

src/commands/distribute/release.ts
```diff
@@ -0,0 +1,224 @@
...
+
+      debug("Finishing release upload");
+      releaseUrl = await this.finishReleaseUpload(client, app, uploadId);
+    } catch (error) {
+      try {
+        out.text("Release upload failed");        
+        await this.abortReleaseUpload(client, app, uploadId);
+        out.text("Release Upload was aborted");
+      } catch (abortError) {
+        out.text("Failed to abort release upload");
```
izikl: This is a really funny message. I don't think we should notify the user about it. 
I would just say keep the original "Release Upload was aborted".

src/commands/distribute/release.ts
```diff
@@ -0,0 +1,224 @@
...
+    debug("Distributing the release");
+    const releaseDetails = await this.distributeRelease(client, app, releaseId, releaseNotesString);    
+
+    out.text(`Release ${releaseDetails.shortVersion} (${releaseDetails.id}) was successfully released to ${distributionGroupUsers.length} testers of ${this.distributionGroup}`);
+
+    return success();
+  }
+
+  private validateParameters(): void {
+    if (_.isNil(this.releaseNotes) && _.isNil(this.releaseNotesFile)) {
```
izikl: release note is optional, having no release notes is a valid flow.

src/commands/distribute/release.ts
```diff
@@ -0,0 +1,224 @@
...
+
+  private async getDistributionGroupUsers(client: MobileCenterClient): Promise<models.DistributionGroupUserGetResponse[]> {
+    const distributionGroupUsersRequestResponse = await clientRequest<models.DistributionGroupUserGetResponse[]>(
+      (cb) => client.account.getDistributionGroupUsers(this.app.ownerName, this.app.appName, this.distributionGroup, cb));
+    const statusCode = distributionGroupUsersRequestResponse.response.statusCode;
+    if (statusCode >= 400) {
+      if (statusCode === 404) {
+        throw failure(ErrorCodes.InvalidParameter, `distribution group ${this.distributionGroup} was not found`);
+      } else {
+        throw failure(ErrorCodes.Exception, `failed to get users of distribution group ${this.distributionGroup}`);
```
izikl: I don't think this should fail the command, I would suggest to just update the message to not show the number of testers. Something like this:
"Release 1.0 (45) was successfully released in Beta Testers"

src/commands/distribute/release.ts
izikl: Well = Will

src/commands/distribute/release.ts
izikl: "in" not "of"

src/commands/distribute/release.ts
bryukov: Yes, it will be.

src/commands/distribute/release.ts
bryukov: Is it OK to make it a debug message? It will not be visible to the user unless debug mode is enable.

src/commands/distribute/release.ts
bryukov: Fixed

src/commands/distribute/release.ts
bryukov: Fixed

src/commands/distribute/release.ts
bryukov: Fixed

src/commands/distribute/release.ts
bryukov: Fixed

bryukov: Fixed

src/commands/distribute/release.ts
bryukov: Fixed

src/commands/distribute/release.ts
izikl: Yes, debug makes sense

src/commands/distribute/release.ts
bryukov: Added `Release ${releaseDetails.shortVersion} (${releaseDetails.version}) was successfully released to ${this.distributionGroup}` output for this case

src/commands/distribute/release.ts
bryukov: Reimplemented `getPrerequisites` to use async/await instead of promises

src/commands/distribute/release.ts
```diff
@@ -0,0 +1,242 @@
...
+  }
+
+  private async doCommand(client: MobileCenterClient): Promise<CommandResult> {
+    const app: DefaultApp = this.app;
+    
+    debug("Check that user hasn't selected both --release-notes and --release-notes-file");
+    this.validateParameters();
+
+    debug("Loading prerequisites");
+    const [distributionGroupUsersNumber, releaseBinaryFileBuffer, releaseNotesString] = await out.progress("Loading prerequisites...", this.getPrerequisites(client));
```
izikl: distributionGroupUsersCount

bryukov: Fixed

src/commands/distribute/release.ts
christav: We're looking at adding the `node-fetch` module, which will give us a promise-native API for doing raw web requests Until then, this is fine.

PR #147, opened 2017-03-23
Implemented "distribute groups show" command

bryukov:
distribute-groups-show
Please note that the expected output was not yet defined when the pull request was created


src/commands/distribute/groups/show.ts
```diff
@@ -0,0 +1,131 @@
...
+          releasesDetails.push(response.result);
+        } else {
+          throw response.response.statusCode;
+        }
+      } catch (error) {
+        debug(`Failed to get release details for release ${basicReleasesDetails[i].id} - ${inspect(error)}`);        
+      }
+    }
+
+    // TODO implement expected output
```
bryukov: It needs to be re-done once the expected output will be specified.
Right now it outputs all the fields of DistributionGroupUserGetResponse and ReleaseDetails. It is also not JSON friendly for now. 
@JoshuaWeber , @izikl , can you, please, provide the example of the expected output for this command?

bryukov: Re-done in accordance to @izikl answers

PR #149, opened 2017-03-24
Implemented 'distribute groups create' command

bryukov:
capture


src/commands/distribute/groups/create.ts
```diff
@@ -0,0 +1,133 @@
...
+  public testers: string;
+
+  @help("Path to file containing list of testers")
+  @shortName("T")
+  @longName("testers-file")
+  @hasArg
+  public testersListFile: string;
+
+  public async run(client: MobileCenterClient): Promise<CommandResult> {
+    try {
```
christav: I've seen this exact code in `run` a couple times now. Is this something we should pull up into the framework so you don't need to repeat that catch clause in every command?

bryukov: Yes, I think, it should be pulled up. I have created the pull request with this change - https://github.com/Microsoft/mobile-center-cli/pull/158

PR #160, opened 2017-04-06
Implemented 'analytics log-flow' command

bryukov:
Implemented 'analytics log-flow' command according to the specification
analytics_log-flow
Continuous Mode can be exited by pressing Ctrl+C


src/commands/analytics/log-flow.ts
```diff
@@ -0,0 +1,147 @@
...
+      return logs.filter((logEntry) => logEntry.installId === installId);
+    } else {
+      return logs;
+    }
+  }
+
+  private showLogEntry(output: StreamingArrayOutput, logEntry: models.Log): void {
+    // setting common properties
+    let logString: string = `${logEntry.timestamp}, ${logEntry.installId}, ${logEntry.type}`;
+    let jsonObject = _({
```
bryukov: Yes, it is not really needed here. I'll change it it.

src/commands/analytics/log-flow.ts
```diff
@@ -0,0 +1,147 @@
...
+      date: logEntry.timestamp,
+      installId: logEntry.installId,
+      logType: logEntry.type
+    });
+
+    // adding log id
+    if (isEventLog(logEntry)) {
+      // event name for event log
+      logString += ", " + logEntry.name;
+      jsonObject.extend({ logId: logEntry.name});
```
bryukov: I have implemented `jsonObject.logId = logEntry.name` approach

src/commands/analytics/log-flow.ts
```diff
@@ -0,0 +1,147 @@
...
+
+    // adding properties
+    if (isLogWithProperties(logEntry)) {
+      if (_.size(logEntry.properties)) {
+        logString += ", " + _.toPairs(logEntry.properties).map((pair) => pair.join(": ")).join(", ");
+      }
+      jsonObject.extend({ properties: logEntry.properties});
+    }
+
+    output.text(_.constant(logString), jsonObject);
```
bryukov: No, `_.constant(logString)` is equal to `() => logString`. I have changed it to the latter.

src/util/continuous-polling/continuous-polling-helper.ts
```diff
@@ -0,0 +1,54 @@
...
+    // cancelling timeout timer
+    clearTimeout(timeoutTimer);
+    // closing readline interface
+    readline.close();
+  }
+}
+
+function waitForCtrlC(readline: Readline.ReadLine): Promise<Symbol> {
+  return new Promise<Symbol>((resolve) => {
+    readline.once("SIGINT", () => {
```
bryukov: Thanks!

src/util/interaction/out.ts
```diff
@@ -356,7 +356,12 @@ report.allProperties = function (data: any): any {
 };
 
 report.asDate = function (data: any): string {
-  return new Date(Date.parse(data)).toString();
+  const date = new Date(data);
+  if (formatIsJson()) {
+    return date.toJSON();
```
bryukov: I have noticed that the example from command spec uses ISO string for JSON. It also seems to be used commonly for Dates inside the JSON.

PR #161, opened 2017-04-07
Implemented "analytics app-versions" command

bryukov:
analytics-app-versions


src/commands/analytics/app-versions.ts
```diff
@@ -0,0 +1,87 @@
...
+import * as Path from "path";
+import * as Pfs from "../../util/misc/promisfied-fs";
+import { DefaultApp } from "../../util/profile";
+import * as Os from "os";
+
+const debug = require("debug")("mobile-center-cli:commands:analytics:app-versions");
+
+@help("Shows versions of the application")
+export default class ShowAppVersionsCommand extends AppCommand {
+  @help("Start date")
```
christav: Include a sample date string in the help so people know the format.

src/commands/analytics/app-versions.ts
```diff
@@ -0,0 +1,87 @@
...
+  @shortName("s")
+  @longName("start")
+  @hasArg
+  public startDate: string;
+
+  @help("End date")
```
christav: Include a sample date string so people know the format.

src/commands/analytics/app-versions.ts
bryukov: Done

src/commands/analytics/app-versions.ts
bryukov: Done

PR #162, opened 2017-04-07
Continuous polling for 'build logs'

bryukov:
Implemented '-c/--continue' option for 'build logs' command


bryukov:
Please, note that this PR depends on #160

PR #163, opened 2017-04-11
Test Cloud commands: adding timeout and fixing "please inspect logs" messages

kmiecikt:
This PR contains two changes:
1. It adds `timeout` argument to all Test Cloud `run` commands. That will allow us to limit the time that we spend waiting for test completion, and is needed for Build - Test integration.
2. Improves messages. 
Many of our users complaint that the message "please inspect logs for more details" is confusing - there are no logs. This message is now removed; instead, the code prints exit code, what can be helpful during automation.

## Sample output for `run` command with `timeout`
```
...
Current test status: Validating
Current test status: Validation completed. Waiting for devices
After 30 seconds, command timed out waiting for tests to finish.
Command failed: cannot run Test Cloud tests. Returning exit code 64.
```

### DIFF
--- a/bin/mobile-center.js
+++ b/bin/mobile-center.js
@@ -36,7 +36,7 @@ function runCli() {
   runner(process.argv.slice(2))
     .then(function (result) {
       if (commandLine.failed(result)) {
-        console.log(`Command failed, ${result.errorMessage}`);
+        console.log(`Command failed: ${result.errorMessage}`);
         process.exit(result.errorCode);
       }
     });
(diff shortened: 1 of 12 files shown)


bin/mobile-center.js
```diff
@@ -36,7 +36,7 @@ function runCli() {
   runner(process.argv.slice(2))
     .then(function (result) {
       if (commandLine.failed(result)) {
-        console.log(`Command failed, ${result.errorMessage}`);
+        console.log(`Command failed: ${result.errorMessage}`);
```
kmiecikt: This is just a proposal: I think error messages look better if the sub-message appears after colon.

jonstoneman: I like the `"Command failed:"` prefix but I think the error messages should begin with capital letters otherwise it looks odd when the messages contain multiple sentences:

`"Command failed: cannot run Test Cloud tests. Returning exit code 64."`

kmiecikt: Done - all the error messages begin with capital letter.

PR #164, opened 2017-04-11
Fixed 'distribute' commands issues which were found during Bug Bash

bryukov:


src/commands/distribute/groups/update.ts
```diff
@@ -159,28 +171,30 @@ export default class UpdateDistributionGroupCommand extends AppCommand {
...
         clientRequest<models.DistributionGroupUserDeleteResponse[]>((cb) => client.account.deleteDistributionGroupUsers(app.ownerName, app.appName, this.distributionGroup, {
           userEmails
         }, cb)));
       if (httpResponse.response.statusCode >= 400) {
         throw httpResponse.response.statusCode;
       } else {
-        return httpResponse.result;
+        // TODO: replace once result swagger is fixed
+        // return httpResponse.result.filter((result) => result.status < 400).map((result) => result.userEmail);
+        return (<any[]> JSON.parse((<any> httpResponse.response).body)).filter((result) => result.status < 400).map((result) => result.user_email);
```
bryukov: There is a problem with the swagger for deleteDistributionGroupUsers operation. It specifies the single object DistributionGroupUserDeleteResponse as a result, but the operation actually returns an array of DistributionGroupUserDeleteResponse objects. 
I have implemented this workaround for now. Once the bug will be fixed, it will be replaced with the previous line.

izikl: can you support both?

bryukov: Discussed offline.

PR #182, opened 2017-05-10
Implemented auto-completion support for categories, commands and options

bryukov:
Implemented auto-completion for categories, commands and options using [Omelette](https://www.npmjs.com/package/omelette).
autocomplete

Auto-completion script can be installed using "setup-autocomplete" command. Bash, zsh and fish shells are supported.
setup-autocomplete

Terminal should be restarted afterwards for the changes to take effect.

### DIFF
--- a/bin/mobile-center.js
+++ b/bin/mobile-center.js
@@ -43,7 +43,13 @@ function runCli() {
 }
 
 if (ensureNodeVersion()) {
-  runCli();
+  if(process.argv.indexOf("--compgen") === -1) {
+    runCli();
+  } else {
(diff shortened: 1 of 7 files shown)


bryukov:
Implemented handling of "help" command

christav:
Looks nice. I wonder - would it be possible to extract all the command line information at build time and have a faster completion path through that wouldn't require loading up the entire CLI to calculate the completion data? Or does that happen under the hood with omelette somehow?

bryukov:
Implemented autocomplete tree file generation during project building to improve autocompletion performance.

PR #187, opened 2017-05-18
Use interactive login from CLI

christav:
Now that we have an oauth flow login page of the server, this PR launches a browser on login to collect user information. This way we can log in with github, AAD, or whatever else the server supports.


src/util/profile/environments.ts
```diff
@@ -20,22 +20,27 @@ const environmentsData: EnvironmentsFile = {
...
       description: "Integration"
     },
     staging: {
       endpoint: "https://bifrost-staging.trafficmanager.net",
+      loginEndpoint: "https://asgard-staging.trafficmanager.net/cli-login",
       description: "Staging"
     },
     prod: {
       endpoint: "https://api.mobile.azure.com",
+      loginEndpoint: "https://mobile.aure/com/cli-login",
```
lumaxis: Typo? 🙂

christav: D'oh!

PR #196, opened 2017-05-30
Implemented "orgs create" command

bryukov:
Implemented command according to the specification
orgs-create


src/util/portal/portal-helper.ts
```diff
@@ -1,11 +1,18 @@
 let portalMap: Map<string, string> = new Map<string, string>([
-  ["https://bifrost-int.trafficmanager.net", "https://asgard-int.trafficmanager.net"]
```
christav: Not really part of this PR, but would it be worth it to move these uris to the environments.ts file to centralize all the endpoint tracking in one place?

PR #205, opened 2017-05-31
Implemented "orgs collaborators list" command

bryukov:
orgs-collaborators-list

### DIFF
--- a/src/commands/orgs/lib/org-users-helper.ts
+++ b/src/commands/orgs/lib/org-users-helper.ts
@@ -1,0 +1,25 @@
+import { MobileCenterClient, models, clientRequest, ClientResponse } from "../../../util/apis";
+import { failure, ErrorCodes } from "../../../util/commandline";
+import { inspect } from "util";
+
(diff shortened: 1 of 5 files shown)


src/commands/orgs/lib/org-users-helper.ts
```diff
@@ -0,0 +1,25 @@
+import { MobileCenterClient, models, clientRequest, ClientResponse } from "../../../util/apis";
+import { failure, ErrorCodes } from "../../../util/commandline";
+import { inspect } from "util";
+
+export async function getOrgUsers(client: MobileCenterClient, organization: string, debug: Function): Promise<models.OrganizationUserResponse[]> {
+  try {
+    const httpResponse = await clientRequest<models.OrganizationUserResponse[]>((cb) => client.users.listForOrg(organization, cb));
+    if (httpResponse.response.statusCode < 400) {
```
christav: I've seen this pattern repeated a lot - do we need to move this logic somehow into clientRequest?

PR #220, opened 2017-06-09
Re-introduced "analytics log-flow" command

bryukov:
"analytics log-flow" command uses forward-compatible analytics.genericLogFlow API method from now


src/commands/analytics/log-flow.ts
```diff
@@ -0,0 +1,138 @@
...
+    streamingOutput.start();
+
+    let options: {start: Date};
+    await ContinuousPollingHelper.pollContinuously(async (requestsDone: number) => {
+      try {
+        debug ("Loading logs");
+        // start time is not specified for the first request
+        return await clientRequest(
+          (cb) => requestsDone ? 
+            client.analytics.genericLogFlow(app.ownerName, app.appName, options, cb) :
```
christav: I'm pretty sure if you leave options `undefined` or `null` it'll act the same as if you didn't pass it at all, so I don't think you need the `requestsDone ?` condition here.

Yeah, I just double checked the client code - just initialize options to `null` the first time through and pass it in, everything will work.

bryukov: Done, thanks!

PR #246, opened 2017-06-23
Progress spinner is now shown during the whole logout process

bryukov:
Updated "logout" command to make sure that the app doesn't always waits for 10 seconds before exiting


christav:
The goal with the 10 second timeout is that sometimes it takes a very long time to delete a token. Rather than make the user wait forever to log out before actually being logged in, the current code means that the users don't wait more than 10 seconds before the process goes on. I would not remove this logic.

PR #255, opened 2017-06-27
Appropriate messages are shown when command has nothing to output

bryukov:
analytics app-versions:
app-versions-empty
analytics log-flow:
log-flow-empty
apps get-current:
get-current-empty
apps list:
apps-list-empty
build branches list:
build-branches-list-empty
distribute releases list:
distribute-releases-list-empty

The following commands now also show messages when they finish:
apps set-current:
apps-set-current
crashes upload-symbols:
crashes-upload-symbols
apps get-current:
apps-get-current-empty

Based on #245


christav:
@bryukov Conflicts, and I disagree with this one - I'm going by the unix command line philosophy of "silence is golden." It's pretty standard if a command has nothing to output it just outputs nothing.

christav:
Obsolete and not agreed - this kind of thing makes it harder to parse the output in automation scenarios. Closing.

PR #257, opened 2017-06-29
Add 'mobile-center test stop' command for stopping the started test run.

dkandrov:
The added command has been implemented in order to support Jenkins Plugin for Mobile Center Test.
Also, users will be able to cancel the testing process from the command line when they want.

### DIFF
--- a/src/commands/test/stop.ts
+++ b/src/commands/test/stop.ts
@@ -17,7 +17,7 @@ export default class StopCommand extends AppCommand {
 
   async run(client: MobileCenterClient): Promise<CommandResult> {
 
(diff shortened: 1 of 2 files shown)


src/commands/test/stop.ts
```diff
@@ -0,0 +1,35 @@
...
+  @hasArg
+  testRunId: string;
+
+  constructor(args: CommandArgs) {
+    super(args);
+  }
+
+  async run(client: MobileCenterClient): Promise<CommandResult> {
+
+    this.stopTestRun(client, this.testRunId);
```
christav: You need to await this promise or things won't clean up/report errors properly.

PR #266, opened 2017-08-03
Codepush - deployment - list

buptkang:
This is the first PR to start integrating codepush cli into mobile center cli. 

Original codepush deployment list command: `code-push deployment ls <app-name>`
screen shot 2017-08-03 at 1 56 06 pm

In the mobile center, for the same command, users can type `mobile-center codepush deployment list` without specifying the app name. 

screen shot 2017-08-03 at 1 53 58 pm

screen shot 2017-08-03 at 1 54 12 pm
screen shot 2017-08-03 at 1 54 28 pm


src/commands/codepush/deployment/list.ts
```diff
@@ -0,0 +1,30 @@
+import { AppCommand, CommandArgs, CommandResult, help, failure, ErrorCodes, success, getCurrentApp } from "../../../util/commandline";
+import { out } from "../../../util/interaction";
+import { DefaultApp } from "../../../util/profile";
+import { inspect } from "util";
+import { MobileCenterClient, models, clientRequest } from "../../../util/apis";
+const _ = require("lodash");
+
+const debug = require("debug")("mobile-center-cli:commands:codepush:deployments:list");
+
+@help("List the deployments associated with an app")
```
pniko: I think this should be changed to "List the deployments associated with your current app" so the user knows its applicable to their current application.

src/commands/codepush/category.txt
```diff
@@ -0,0 +1 @@
+codepush in mobile center
```
pniko: This should follow the convention utilized elsewhere for other commands. "View and manage CodePush deployments and releases" would be a better option.

src/commands/codepush/deployment/list.ts
```diff
@@ -0,0 +1,30 @@
...
+export default class ListCommand extends AppCommand {
+  constructor(args: CommandArgs) {
+    super(args);
+  }
+
+  async run(client: MobileCenterClient): Promise<CommandResult> {
+    const app = this.app;
+    let deployments: models.Deployment[];
+    try {
+      const httpRequest = await out.progress("Getting codepush deployments ...", clientRequest<models.Deployment[]>(
```
pniko: Remove space before ellipsis (...) on the end.

src/commands/codepush/deployment/ls.ts
```diff
@@ -0,0 +1,30 @@
...
+import { out } from "../../../util/interaction";
+import { DefaultApp } from "../../../util/profile";
+import { inspect } from "util";
+import { MobileCenterClient, models, clientRequest } from "../../../util/apis";
+const _ = require("lodash");
+
+const debug = require("debug")("mobile-center-cli:commands:codepush:deployments:ls");
+
+@help("List the deployments associated with an app")
+export default class LsCommand extends AppCommand {
```
pniko: You are duplicating all this logic between two classes. At minimum we need to remove one command or the other, at most you should extract the logic into a standard base class. I would recommend removing 'ls' and only keep 'list' so it matches the rest of the CLI naming conventions.

src/commands/codepush/deployment/list.ts
christav: That's assumed on every command. Also, what it they used the `--app` switch instead of `apps set-current`?

src/commands/codepush/deployment/list.ts
```diff
@@ -0,0 +1,30 @@
...
+        (cb) => client.deployments.list(app.ownerName, app.appName, cb)));
+      deployments = httpRequest.result;
+    } catch (error) {
+      debug(`Failed to get list of codepush deployments - ${inspect(error)}`);
+      return failure(ErrorCodes.Exception, "failed to get list of deployments for the app");
+    }
+    out.table(out.getCommandOutputTableOptions(["Name", "Key"]), deployments.map((deployment) => [deployment.name, deployment.key]));
```
christav: I'd move these two lines inside the try block - they don't get executed on exception anyway, and this way you don't need the function level scoped "deployments" variable, you can bring that inside as well.

Also, since you're not using anything but the result of the http request (not looking at response headers for example) you can use `clientCall` instead of `clientRequest`. The return value of `clientCall` is the result of the call, so you don't have to manage it yourself like you're doing here.

src/commands/codepush/deployment/ls.ts
christav: Why are there two files here at all? They call the same back end and display the results in the same format. If there's a difference between the two that I missed it's probably better to vary with an option rather than a completely separate command.

If it's just to get both `list` and `ls` to do the same thing, don't do that. We use `list` everywhere in the CLI, we don't use `ls` for anything.

src/commands/codepush/category.txt
buptkang: Accept this change.

src/commands/codepush/deployment/list.ts
buptkang: Fix the comment, provide better error message to help users to use --app command to specify the app.

src/commands/codepush/deployment/list.ts
```diff
@@ -0,0 +1,36 @@
...
+    const app = this.app;
+    let deployments: models.Deployment[];
+    try {
+      const httpRequest = await out.progress("Getting codepush deployments...", clientRequest<models.Deployment[]>(
+        (cb) => client.deployments.list(app.ownerName, app.appName, cb)));
+      deployments = httpRequest.result;
+      out.table(out.getCommandOutputTableOptions(["Name", "Key"]), deployments.map((deployment) => [deployment.name, deployment.key]));
+      return success();
+    } catch (error) {
+      debug(`Failed to get list of codepush deployments - ${inspect(error)}`);
```
pniko: codepush needs be capitalized.

src/commands/codepush/deployment/list.ts
```diff
@@ -0,0 +1,36 @@
...
+export default class ListCommand extends AppCommand {
+  constructor(args: CommandArgs) {
+    super(args);
+  }
+
+  async run(client: MobileCenterClient): Promise<CommandResult> {
+    const app = this.app;
+    let deployments: models.Deployment[];
+    try {
+      const httpRequest = await out.progress("Getting codepush deployments...", clientRequest<models.Deployment[]>(
```
buptkang: Fixed.

src/commands/codepush/deployment/list.ts
buptkang: Fixed.

src/commands/codepush/deployment/list.ts
```diff
@@ -0,0 +1,36 @@
...
+    try {
+      const httpRequest = await out.progress("Getting Codepush deployments...", clientRequest<models.Deployment[]>(
+        (cb) => client.deployments.list(app.ownerName, app.appName, cb)));
+      deployments = httpRequest.result;
+      out.table(out.getCommandOutputTableOptions(["Name", "Key"]), deployments.map((deployment) => [deployment.name, deployment.key]));
+      return success();
+    } catch (error) {
+      debug(`Failed to get list of Codepush deployments - ${inspect(error)}`);
+      if (error.statusCode === 404) {
+        const appNotFoundErrorMsg = "The app name does not exist. Please double check the name, and provide it in the form owner/appname. \nRun the command " + chalk.bold("mobile-center apps list") + " to see what apps you have access to.";
```
christav: Instead of saying "The app name does not exist" in the message, actually put the app name the user typed in there, so they will obviously see if they made a typo.

christav: Also consider using template string syntax to avoid a lot of string concatenations.

buptkang: @christav That is a good point to let users notice what they mistyped. I have updated the code.

src/commands/codepush/deployment/list.ts
```diff
@@ -26,9 +26,9 @@ export default class ListCommand extends AppCommand {
     } catch (error) {
       debug(`Failed to get list of Codepush deployments - ${inspect(error)}`);
       if (error.statusCode === 404) {
-        const appNotFoundErrorMsg = "The app name does not exist. Please double check the name, and provide it in the form owner/appname. \nRun the command " + chalk.bold("mobile-center apps list") + " to see what apps you have access to.";
+        const appNotFoundErrorMsg = `The app ${app.ownerName}\\${app.appName} does not exist. Please double check the name, and provide it in the form owner/appname. \nRun the command ${chalk.bold("mobile-center apps list")} to see what apps you have access to.`;
```
christav: Mobile center uses forward slashes (`/`) not backslashes (`\`) to separate owner and app. Fix this.

buptkang: Fixed.

PR #270, opened 2017-08-08
code-push-patch

buptkang:
This PR migrates `codepush patch` into mobile center:

The original codepush patch command [illustration](https://github.com/Microsoft/code-push/tree/master/cli#patching-update-metadata).

The current version:
screen shot 2017-08-08 at 3 33 38 pm

### DIFF
--- a/src/commands/codepush/patch.ts
+++ b/src/commands/codepush/patch.ts
@@ -2,7 +2,6 @@ import { AppCommand, CommandArgs, CommandResult, help, failure, ErrorCodes, succ
 import { out } from "../../util/interaction";
 import { inspect } from "util";
 import { MobileCenterClient, models, clientRequest, clientCall } from "../../util/apis";
-import * as _ from "lodash";
 import * as chalk from "chalk";
 import * as semver from "semver";
 
@@ -62,14 +61,14 @@ export default class PatchCommand extends AppCommand {
       return failure(ErrorCodes.Exception, "At least one property must be specified to patch a release.");
     }
 
-    if (this.rollout != null && this.rollout !== undefined) {
-      if (parseInt(this.rollout) < 0 || parseInt(this.rollout) > 100 || !/^(100|[1-9][0-9]|[1-9])$/.test(this.rollout)) {
+    if (this.rollout != null) {
+      if (!/^(100|[1-9][0-9]|[1-9])$/.test(this.rollout)) {
         return failure(ErrorCodes.Exception, `Rollout value should be integer value between ${chalk.bold('0')} or ${chalk.bold('100')}.`);
       }
     }
 
     const isValidVersion = (version: string): boolean => !!semver.valid(version) || /^\d+\.\d+$/.test(version) || /^\d+$/.test(version);
-    if (this.targetBinaryRange !== null && this.targetBinaryRange !== undefined && !isValidVersion(this.targetBinaryRange)) {
+    if (this.targetBinaryRange != null && !isValidVersion(this.targetBinaryRange)) {
       return failure(ErrorCodes.Exception, "Invalid binary version(s) for a release.");
     }
 
--- a/src/commands/codepush/patch.ts
+++ b/src/commands/codepush/patch.ts
@@ -90,14 +90,24 @@ export default class PatchCommand extends AppCommand {
       return failure(ErrorCodes.Exception, "Invalid binary version(s) for a release.");      
     }
 
-    let patch: models.LiveUpdateReleaseModification = {
-      targetBinaryRange: this.targetBinaryRange,
-      isMandatory: this.isMandatory === 'true' || this.isMandatory === 'True',
-      isDisabled: this.isDisabled === 'true' || this.isDisabled === 'True',
-      description: this.description,
-      rollout: parseInt(this.rollout)
+    let patch: models.LiveUpdateReleaseModification;
+    if (this.rollout!=null) {
+      patch = {
+        targetBinaryRange: this.targetBinaryRange,
+        isMandatory: this.isMandatory === 'true' || this.isMandatory === 'True',
+        isDisabled: this.isDisabled === 'true' || this.isDisabled === 'True',
+        description: this.description,
+        rollout: parseInt(this.rollout)
+      }
+    } else {
+      patch = {
+        targetBinaryRange: this.targetBinaryRange,
+        isMandatory: this.isMandatory === 'true' || this.isMandatory === 'True',
+        isDisabled: this.isDisabled === 'true' || this.isDisabled === 'True',
+        description: this.description,
+      }
     }
-
+    
     try {
       const httpRequest = await out.progress("Patching CodePush release...", clientRequest<models.LiveUpdateRelease>(
         (cb) => client.deploymentReleases.update(this.deploymentName, this.releaseLabel, patch, app.ownerName, app.appName, cb)));
--- a/src/commands/codepush/patch.ts
+++ b/src/commands/codepush/patch.ts
@@ -1,0 +1,116 @@
+import { AppCommand, CommandArgs, CommandResult, help, failure, ErrorCodes, success, getCurrentApp, shortName, longName, required, hasArg, position, name } from "../../util/commandline";
+import { out } from "../../util/interaction";
+import { DefaultApp } from "../../util/profile";
+import { inspect } from "util";
+import { MobileCenterClient, models, clientRequest, clientCall } from "../../util/apis";
+const _ = require("lodash");
+const chalk = require("chalk");
+import * as semver from "semver";
+
+const debug = require("debug")("mobile-center-cli:commands:codepush:patch");
+
+@help("Update the metadata for an existing release")
+export default class PatchCommand extends AppCommand {
+  
+  @help("CodePush deployment name")
+  @required
+  @name("ExistingDeploymentName")
+  @position(0)
+  public deploymentName: string;
+
+  @help("CodePush release label")
+  @required
+  @name("ExistingReleaseLabel")
+  @position(1)
+  public releaseLabel: string;
+
+  @help("update whether the release should be considered mandatory or not")
+  @shortName("m")
+  @longName("mandatory")
+  @hasArg
+  public isMandatory: string;
+
+  @help("update the description associated with the release")
+  @shortName("x")
+  @longName("disabled")
+  @hasArg
+  public isDisabled: string;
+
+  @help("update the semver range that indicates which binary version(s) a release")
+  @shortName("t")
+  @longName("targetBinaryVersion")
+  @hasArg
+  public targetBinaryRange: string;
+
+  @help("update the description associated with the release")
+  @shortName("des")
+  @longName("description")
+  @hasArg
+  public description: string;
+
+  @help("allows you to increase the rollout percentage of the target release")
+  @shortName("r")
+  @longName("rollout")
+  @hasArg
+  public rollout: string;
+
+  constructor(args: CommandArgs) {
+    super(args);
+  }
+
+  async run(client: MobileCenterClient): Promise<CommandResult> {
+
+    const app = this.app;
+    let release: models.LiveUpdateRelease;
+
+    if (this.targetBinaryRange == null && this.isDisabled == null && this.isMandatory == null && this.description == null && this.rollout == null) {
+      return failure(ErrorCodes.Exception, "At least one property must be specified to patch a release.");
+    }
+
+    if (this.isMandatory != null) {
+      if (this.isMandatory != 'true' && this.isMandatory != 'false') {
+        return failure(ErrorCodes.Exception, `Mandatory value should be either ${chalk.bold(true)} or ${chalk.bold(false)}.`);  
+      }
+    }
+
+    if (this.isDisabled != null) {
+      if (this.isDisabled != 'true' && this.isDisabled != 'false') {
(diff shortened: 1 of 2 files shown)


src/commands/codepush/patch.ts
```diff
@@ -0,0 +1,126 @@
+import { AppCommand, CommandArgs, CommandResult, help, failure, ErrorCodes, success, getCurrentApp, shortName, longName, required, hasArg, position, name } from "../../util/commandline";
+import { out } from "../../util/interaction";
+import { DefaultApp } from "../../util/profile";
```
sergey-akhalkov: Unused `DefaultApp` import here, I believe you should remove all unused imports if any.

src/commands/codepush/patch.ts
```diff
@@ -0,0 +1,126 @@
...
+      }
+    }
+
+    const isValidVersion = (version: string): boolean => !!semver.valid(version) || /^\d+\.\d+$/.test(version);
+    if (this.targetBinaryRange != null && !isValidVersion(this.targetBinaryRange)) {
+      return failure(ErrorCodes.Exception, "Invalid binary version(s) for a release.");      
+    }
+
+    let patch: models.LiveUpdateReleaseModification;
+    if (this.rollout!=null) {
```
sergey-akhalkov: Code style issue here, should be `this.rollout != null`

src/commands/codepush/patch.ts
```diff
@@ -0,0 +1,126 @@
...
+      patch = {
```
sergey-akhalkov: Duplicate code is here, what do you think about this?
```js
patch = {
    targetBinaryRange: this.targetBinaryRange,
    isMandatory: this.isMandatory === 'true' || this.isMandatory === 'True',
    isDisabled: this.isDisabled === 'true' || this.isDisabled === 'True',
    description: this.description,
}

if (this.rollout != null) {
    patch.rollout = parseInt(this.rollout);
}
```

src/commands/codepush/patch.ts
```diff
@@ -0,0 +1,126 @@
...
+        targetBinaryRange: this.targetBinaryRange,
+        isMandatory: this.isMandatory === 'true' || this.isMandatory === 'True',
```
sergey-akhalkov: `isMandatory: this.isMandatory === 'true' || this.isMandatory === 'True',` - I believe it would be better to use/impelent boolean variables parser/validator to avoid such kind of expressions.

src/commands/codepush/patch.ts
```diff
@@ -0,0 +1,126 @@
...
+      }
+    }
+
+    if (this.rollout != null) {
+      if (parseInt(this.rollout) < 0 || parseInt(this.rollout) > 100) {
+        return failure(ErrorCodes.Exception, `Rollout value should be integer value between ${chalk.bold(0)} or ${chalk.bold(100)}.`);
+      }
+    }
+
+    const isValidVersion = (version: string): boolean => !!semver.valid(version) || /^\d+\.\d+$/.test(version);
```
sergey-akhalkov: `const isValidVersion` check is out of date here, should be `const isValidVersion = (version: string): boolean => !!semver.valid(version) || /^\d+\.\d+$/.test(version) || /^\d+$/.test(version);`
Please take a look at https://github.com/Microsoft/code-push/blob/c5a294d503226536b3022b9e0c86303c2d14e29d/cli/script/command-executor.ts#L902

src/commands/codepush/patch.ts
```diff
@@ -0,0 +1,126 @@
...
+    }
+
+    if (this.isDisabled != null) {
+      if (this.isDisabled != 'true' && this.isDisabled != 'false') {
+        return failure(ErrorCodes.Exception, `Disabled value should be either ${chalk.bold(true)} or ${chalk.bold(false)}.`);  
+      }
+    }
+
+    if (this.rollout != null) {
+      if (parseInt(this.rollout) < 0 || parseInt(this.rollout) > 100) {
```
sergey-akhalkov: Should not we validate if `this.rollout` is a number or not?
Probably like this 
```js
export function isValidRolloutField(rollout: any): boolean {
        // rollout is an optional field, or when defined should be a number between 1-100.
        return /^(100|[1-9][0-9]|[1-9])$/.test(rollout);
    }
```

src/commands/codepush/patch.ts
```diff
@@ -0,0 +1,126 @@
+import { AppCommand, CommandArgs, CommandResult, help, failure, ErrorCodes, success, getCurrentApp, shortName, longName, required, hasArg, position, name } from "../../util/commandline";
+import { out } from "../../util/interaction";
+import { DefaultApp } from "../../util/profile";
+import { inspect } from "util";
+import { MobileCenterClient, models, clientRequest, clientCall } from "../../util/apis";
+const _ = require("lodash");
```
christav: Use `import` not `require`

src/commands/codepush/patch.ts
```diff
@@ -0,0 +1,126 @@
...
+  public isDisabled: string;
+
+  @help("update the semver range that indicates which binary version(s) a release")
+  @shortName("t")
+  @longName("targetBinaryVersion")
+  @hasArg
+  public targetBinaryRange: string;
+
+  @help("update the description associated with the release")
+  @shortName("des")
```
christav: shortName *must* be one character and one character only.

`mobile-center thing -des` will be parsed as if it was `mobile-center thing -d -e -s`

src/commands/codepush/patch.ts
```diff
@@ -0,0 +1,126 @@
...
+  @longName("description")
+  @hasArg
+  public description: string;
+
+  @help("allows you to increase the rollout percentage of the target release")
```
christav: This description needs work. HOW does it increase the rollout percentage? Tell me what it does, not what it lets me do (if that makes any sense).

src/commands/codepush/patch.ts
```diff
@@ -0,0 +1,126 @@
...
+  constructor(args: CommandArgs) {
+    super(args);
+  }
+
+  async run(client: MobileCenterClient): Promise<CommandResult> {
+
+    const app = this.app;
+    let release: models.LiveUpdateRelease;
+
+    if (this.targetBinaryRange == null && this.isDisabled == null && this.isMandatory == null && this.description == null && this.rollout == null) {
```
christav: Always use `===`, not `==` in Javascript, particularly when comparing to null.

src/commands/codepush/patch.ts
```diff
@@ -0,0 +1,126 @@
...
+      return failure(ErrorCodes.Exception, "At least one property must be specified to patch a release.");
+    }
+
+    if (this.isMandatory != null) {
```
christav: And use `!==`, not `!=`

src/commands/codepush/patch.ts
```diff
@@ -0,0 +1,126 @@
...
+      if (this.isMandatory != 'true' && this.isMandatory != 'false') {
```
christav: Would it be easier for users if this is a boolean flag instead of a true/false string? By that, I mean assume that it's not mandatory, and if they want it mandatory then specify the `-m` flag. Or the reverse, whatever makes sense for your scenario. But it strikes me that this would be better than having to explicitly type out `true` and `false`.

src/commands/codepush/patch.ts
```diff
@@ -0,0 +1,126 @@
...
+        return failure(ErrorCodes.Exception, `Mandatory value should be either ${chalk.bold(true)} or ${chalk.bold(false)}.`);  
+      }
+    }
+
+    if (this.isDisabled != null) {
```
christav: Same comment about boolean flags here.

src/commands/codepush/patch.ts
christav: Make it a switch flag instead of an argument and then the infrastructure would handle the parsing & conversion for you.

src/commands/codepush/patch.ts
```diff
@@ -0,0 +1,102 @@
...
+
+    const app = this.app;
+    let release: models.LiveUpdateRelease;
+
+    if (this.targetBinaryRange === null && this.isDisabled === null && this.isMandatory === null && this.description === null && this.rollout === null) {
+      return failure(ErrorCodes.Exception, "At least one property must be specified to patch a release.");
+    }
+
+    if (this.rollout != null && this.rollout !== undefined) {
+      if (parseInt(this.rollout) < 0 || parseInt(this.rollout) > 100 || !/^(100|[1-9][0-9]|[1-9])$/.test(this.rollout)) {
```
sergey-akhalkov: There is no need to use this `parseInt(this.rollout) < 0 || parseInt(this.rollout) > 100` code due to we already have `!/^(100|[1-9][0-9]|[1-9])$/.test(this.rollout)`

src/commands/codepush/patch.ts
```diff
@@ -0,0 +1,102 @@
+import { AppCommand, CommandArgs, CommandResult, help, failure, ErrorCodes, success, getCurrentApp, shortName, longName, required, hasArg, position, name } from "../../util/commandline";
+import { out } from "../../util/interaction";
+import { inspect } from "util";
+import { MobileCenterClient, models, clientRequest, clientCall } from "../../util/apis";
+import * as _ from "lodash";
```
sergey-akhalkov: Unused import here, need to remove `import * as _ from "lodash";`

src/commands/codepush/patch.ts
```diff
@@ -0,0 +1,102 @@
...
+  async run(client: MobileCenterClient): Promise<CommandResult> {
+
+    const app = this.app;
+    let release: models.LiveUpdateRelease;
+
+    if (this.targetBinaryRange === null && this.isDisabled === null && this.isMandatory === null && this.description === null && this.rollout === null) {
+      return failure(ErrorCodes.Exception, "At least one property must be specified to patch a release.");
+    }
+
+    if (this.rollout != null && this.rollout !== undefined) {
```
sergey-akhalkov: `this.rollout != null` this condition already includes check for undefined, so there is no need use `&& this.rollout !== undefined`

src/commands/codepush/patch.ts
buptkang: Fixed.

src/commands/codepush/patch.ts
buptkang: Fixed.

src/commands/codepush/patch.ts
buptkang: @christav I changed back to the double equal syntax since only using triple equal syntax without checking undefined will cause the issue. (default command without rollout parameter)

src/commands/codepush/patch.ts
```diff
@@ -0,0 +1,101 @@
...
+import * as semver from "semver";
+
+const debug = require("debug")("mobile-center-cli:commands:codepush:patch");
+
+@help("Update the metadata for an existing release")
+export default class PatchCommand extends AppCommand {
+  
+  @help("CodePush deployment name.")
+  @required
+  @name("ExistingDeploymentName")
```
christav: Typical CLI conventions use all lower cases and dashes rather than PascalCase. So `--existing-deployment-name` not `--ExistingDeploymentName`

src/commands/codepush/patch.ts
```diff
@@ -0,0 +1,101 @@
...
+  @shortName("m")
+  public isMandatory: boolean;
+
+  @help("Specifies whether this release should be immediately downloadable. Putting -x flag means disabled.")
+  @shortName("x")
+  public isDisabled: boolean;
+
+  @help("Semver expression that specifies the binary app version(s) this release is targeting (e.g. 1.1.0, ~1.2.3).")
+  @shortName("t")
+  @longName("targetBinaryVersion")
```
christav: I'd prefer if you went to the kebab-case version, but if not at least be consistent on using either initial capital letters or lowercase.

src/commands/codepush/patch.ts
```diff
@@ -0,0 +1,101 @@
...
+  @hasArg
+  public targetBinaryRange: string;
+
+  @help("Description of the changes made to the app with this release.")
+  @shortName("d")
+  @longName("description")
+  @hasArg
+  public description: string;
+
+  @help("Percentage of users this release should be immediately available to. This attribute can only be increased from the current value.")
```
christav: How do they know what the current value is? Is that something set in the portal or some other way? I think this help message either needs to be reworked to explain things better. Or leave out the validation details completely.

src/commands/codepush/patch.ts
buptkang: fixed.

src/commands/codepush/patch.ts
buptkang: fixed.

src/commands/codepush/patch.ts
buptkang: fixed.

src/commands/codepush/patch.ts
```diff
@@ -0,0 +1,101 @@
...
+import * as semver from "semver";
+
+const debug = require("debug")("mobile-center-cli:commands:codepush:patch");
+
+@help("Update the metadata for an existing CodePush release")
+export default class PatchCommand extends AppCommand {
+  
+  @help("Specifies one existing deployment name.")
+  @required
+  @name("--existing-deployment-name")
```
christav: You don't actually include the `--` in the command name, sorry. This results in the flag `----existing-deployment-name` in the cli which I don't think will actually work.

Remove the leading `--`.

PR #279, opened 2017-09-29
Test cloud feature complete plus minor bug fixes

owenniblock:
# Motivation
There is a discrepancy between the parameters available when uploading tests via `test-cloud.exe` and those uploaded by `mobile-center-cli`.

This PR adresses that by introducing the ability to add those parameters to the uitest command.

Also including minor bug fixes with this PR that were bundled up in the assigned work.

# Requirements
This PR won't start functioning until UITest has implemented [this PR](https://github.com/xamarin/Xamarin.UITest/pull/1192) - after which I advise we update the minimum version on UITest within the mobile-center-cli.

This work is part of this [VSTS Ticket](https://msmobilecenter.visualstudio.com/Mobile-Center/_workitems/edit/22164).

And along with the UITest changes resolves [this github issue](https://github.com/Microsoft/mobile-center-cli/issues/237).

# Additional Bug Fixes
- [Test upload fails if .app name contains a space](https://msmobilecenter.visualstudio.com/Mobile-Center/_workitems/edit/22798)
- [Reports test run url to user on successful and unsuccessful test run](https://msmobilecenter.visualstudio.com/Mobile-Center/_workitems/edit/22133)
- [404 error when user/org/app name incorrect](https://msmobilecenter.visualstudio.com/Mobile-Center/_workitems/edit/22799)
- [Better error message for 500 error and other unhandled exceptions](https://msmobilecenter.visualstudio.com/Mobile-Center/_workitems/edit/22132)
- Adds better error messages when sending UITest tests if the specified uitest-tools-dir is not found or the test-cloud.exe is not there.

### DIFF
--- a/src/commands/orgs/list.ts
+++ b/src/commands/orgs/list.ts
@@ -10,7 +10,7 @@ import { getOrgsNamesList } from "./lib/org-users-helper";
 export default class OrgListCommand extends Command {
   async run(client: MobileCenterClient): Promise<CommandResult> {
     // every user is a collaborator of it's own group and of zero or more external groups
-    const orgs = await out.progress("Loading list of organizations...", getOrgsNamesList(client, debug));
+    const orgs = await out.progress("Loading list of organizations...", getOrgsNamesList(client));
     const table = orgs.map((names) => [names.displayName, names.name, names.origin]);
 
     out.table(out.getCommandOutputTableOptions(["Display Name", "Name", "Origin"]), table);
--- a/src/commands/orgs/list.ts
+++ b/src/commands/orgs/list.ts
@@ -4,40 +4,17 @@ import { MobileCenterClient, models, clientRequest } from "../../util/apis";
 
 const debug = require("debug")("mobile-center-cli:commands:orgs:list");
 import { inspect } from "util";
+import { getOrgsNamesList } from "./lib/org-users-helper";
 
 @help("Lists organizations in which current user is collaborator")
 export default class OrgListCommand extends Command {
   async run(client: MobileCenterClient): Promise<CommandResult> {
     // every user is a collaborator of it's own group and of zero or more external groups
-    const orgs = await out.progress("Loading list of organizations...", this.getOrgsNamesList(client));
+    const orgs = await out.progress("Loading list of organizations...", getOrgsNamesList(client, debug));
     const table = orgs.map((names) => [names.displayName, names.name, names.origin]);
 
     out.table(out.getCommandOutputTableOptions(["Display Name", "Name", "Origin"]), table);
 
     return success();
   }
-
-  private async getOrgsNamesList(client: MobileCenterClient): Promise<IEntity[]> {
-    try {
-      const httpResponse = await clientRequest<models.OrganizationResponse[]>((cb) => client.organizations.list(cb));
-      if (httpResponse.response.statusCode < 400) {
-        return httpResponse.result.map((org) => ({
-          name: org.name,
-          displayName: org.displayName,
-          origin: org.origin
-        }));
-      } else {
-        throw httpResponse.response;
-      }
-    } catch (error) {
-      debug(`Failed to load list of organization for current user - ${inspect(error)}`);
-      throw failure(ErrorCodes.Exception, "failed to load list of organization for the user");
-    }
-  }
-}
-
-interface IEntity {
-  name: string;
-  displayName: string;
-  origin: string;
 }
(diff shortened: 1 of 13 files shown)


owenniblock:
I think I'm done on this PR. Let me know if you need any changes and I'll sort them out ASAP.

jonstoneman:
Code looks fine, but...   I've not looked at this project in a very long time and it's been a while since I've done any typescript either.  I've also not tried using it so would advise the command is tested with all options (if not done already) and that someone from the CLI team also reviews.

owenniblock:
I was hoping to create some sort of unit tests / integration tests or if needed a test script (bash or rake or whatever) to test all this but I couldn't find anything already in place. Are there CI scripts which run? Perhaps we could hook in to that process and add some extra tests for the new bits?

owenniblock:
(I have tested manually myself but I'd rather put something more concrete in place - and get someone else testing it before releasing it to the wild as well)

owenniblock:
I've just been looking through the Rakefile in `test-cloud-test-apps` and I could create tests for all this in there. However - what we really need is a way to release and test changes as part of a CI process and then we should run integration scripts on this `beta` to validate those changes before releasing them to the wild. 

I would suggest that rather than merging with `master` we merge with a `develop` branch and then tag the release with `beta` [like this](https://coderwall.com/p/l_7acq/publish-beta-versions-of-npm-modules).

Obviously - I can't see the release process and CI systems at the moment - so there could be stuff like that in place already! If there is an I've missed it, please point me in the right direction :)

src/commands/orgs/list.ts
```diff
@@ -4,40 +4,17 @@ import { MobileCenterClient, models, clientRequest } from "../../util/apis";
...
 const debug = require("debug")("mobile-center-cli:commands:orgs:list");
 import { inspect } from "util";
+import { getOrgsNamesList } from "./lib/org-users-helper";
 
 @help("Lists organizations in which current user is collaborator")
 export default class OrgListCommand extends Command {
   async run(client: MobileCenterClient): Promise<CommandResult> {
     // every user is a collaborator of it's own group and of zero or more external groups
-    const orgs = await out.progress("Loading list of organizations...", this.getOrgsNamesList(client));
+    const orgs = await out.progress("Loading list of organizations...", getOrgsNamesList(client, debug));
```
christav: Debug is used per file - if you pass it around you potentially screw up the logging since we don't know what file the log is coming from anymore. Just use the debug const locally in the function you're calling.

src/commands/orgs/list.ts
```diff
@@ -4,40 +4,17 @@ import { MobileCenterClient, models, clientRequest } from "../../util/apis";
...
     const table = orgs.map((names) => [names.displayName, names.name, names.origin]);
 
     out.table(out.getCommandOutputTableOptions(["Display Name", "Name", "Origin"]), table);
 
     return success();
   }
-
```
christav: Why did you make this refactoring - is it part of the feature you're implementing somehow (seems odd that a testcloud command would need to touch the orgs commands)?

christav:
@owenniblock I took a quick look finally - sorry for the delay. In general it looks fine other than the mixed changes which I mentioned in my review.

We don't have any real CI infrastructure around the CLI at the moment; the theory was always that it was up to the beacon teams to test their commands in whatever way they see fit. As such we don't have a "develop" branch for baking stuff for testing - the assumption is that you guys test your stuff to your satisfaction and we merge to master and release when you think it's ready.

So you guys tell me - are you ready to have this put into master and go into the next release, or do you want to leave it in PR for the moment while you do more testing?

src/commands/orgs/list.ts
owenniblock: This was after some discussions with @skovsboll - I needed to determine if the current run was from a user or an organization (so that I could generate and report the correct URL to the user). Unfortunately there seemed to be no easy way to do this so this was the best option we came up with :)

I moved the list so it was accessible to this new functionality.

Happy to change it if there's a better way!

owenniblock:
@christav - I've updated that `debug` issue. Apologies - I wasn't following through what that was doing properly.

Hopefully I have also answered your other query but message me on Slack if that's not the case.

I'll discuss whether we're ready to go or not in the stand-up tomorrow morning and let you know. I'd feel happier if someone other than myself was testing it as well :) perhaps I can do something with our rake stuff so we can automate our testing on this side before you get to see the PR.

owenniblock:
@christav - I found a strange bug while testing using the `rake` files in `test-cloud-test-apps` - I don't think it's due to anything I've done but checking for null on the state and state.message seems to resolve the issue.

I believe this PR is now ready to go. Do we need to give another team a heads-up that I've refactored that `list.ts` file? They may want to do their own testing.

christav:
@owenniblock The "another" team in question is mine, so consider them informed. :-)

PR #294, opened 2017-11-07
Fix/single fixture error

owenniblock:
## Motivation
Running `uitest` tests with a single `--fixture`, `--include-category` or `--exclude-category` fails because the parameter resolves as a single string rather than an array.

## Fix
I refactored the code to fix this problem for `testParameters` and used the same approach.

I've also upped the minimum UITest version (because older versions don't support these new params) and added a bit of extra information in the failure error.


src/commands/test/lib/run-tests-command.ts
```diff
@@ -77,23 +77,24 @@ export abstract class RunTestsCommand extends AppCommand {
...
+    this.include = this.fixArrayParameter(this.include);
+    
+    if (this.timeoutSec && typeof this.timeoutSec === "string") {
+      this.timeoutSec = parseInt(this.timeoutSec);
     }
+  }
 
-    if (!this.include) {
-      this.include = [];
+  protected fixArrayParameter(input: any): string[]
```
christav: Nitpick: You can replace this function with a single call to `_.flatten([input])`. Might make things simpler.

owenniblock: When I get a mo I'll make that change. Best to keep it tidy! Thanks :)

PR #299, opened 2017-11-09
add "codepush release/release-react/release-cordova" commands

sergey-akhalkov:


src/commands/codepush/lib/cordova-utils.ts
```diff
@@ -0,0 +1,49 @@
+import * as which from "which";
+import * as xml2js from "xml2js";
+import * as fs from "fs";
+import * as path from "path";
+
+export function getCordovaProjectAppVersion(): Promise<string> {
+  return new Promise<string>((resolve, reject) => {
+    try {
+      const projectRoot: string = process.cwd();
```
christav: Seems odd to hard-code this to the current directory, don't you want to give users the option of where to get this from?

src/commands/codepush/lib/file-utils.ts
```diff
@@ -10,11 +11,11 @@ export function isBinaryOrZip(path: string): boolean {
 
 export async function copyFileToTmpDir(filePath: string): Promise<string> {
   if (!(await pfs.stat(filePath)).isDirectory()) {
-    var outputFolderPath: string = await pfs.mkTempDir("code-push");
-    await pfs.rmDir(outputFilePath)
+    let outputFolderPath: string = await pfs.mkTempDir("code-push");
```
christav: You are allowed to use sync versions of file system stuff and skip the promises in the CLI.

src/util/commandline/option-decorators.ts
```diff
@@ -181,7 +181,7 @@ export function required(proto: any, propertyKey: string): void {
 
 // DefaultValue is also special, since it has to work with both as well. Same
 // basic logic
-export function defaultValue(value: string | number): PropertyDecorator {
+export function defaultValue(value: string): PropertyDecorator {
```
christav: Didn't you just add the `| number` in your last PR?

sergey-akhalkov: @christav, yes, we've added the `| number` in the https://github.com/Microsoft/mobile-center-cli/pull/293
The main idea was to support `number` type for args in addition to `string` and `boolean`, but unfortunately it does not work like that.

src/commands/codepush/lib/cordova-utils.ts
sergey-akhalkov: The `codepush release-react/cordova` commands are designed to be executed in the root dirs of the projects we are wanting to release updates.
But Agreed, we can update the method signature to make it possible to specify the `projectRoot`. We'll prepare another one PR with several fixes/improvements for release commands.

src/commands/codepush/lib/file-utils.ts
sergey-akhalkov: Got it, will fix in the next PR

PR #300, opened 2017-11-10
additional improvements for "codepush release" related commands

sergey-akhalkov:
1. add missed `hasArg`s
2. add `projectRoot?` prop for `getCordovaProjectAppVersion` and `getReactNativeProjectAppVersion` methods
3. make all method in `file-utils.ts` synchronous
4. add platform validation both for React and Cordova
5. other small fixes and improvements

### DIFF
--- a/src/commands/codepush/lib/cordova-utils.ts
+++ b/src/commands/codepush/lib/cordova-utils.ts
@@ -3,10 +3,10 @@ import * as xml2js from "xml2js";
 import * as fs from "fs";
 import * as path from "path";
 
-export function getCordovaProjectAppVersion(): Promise<string> {
+export function getCordovaProjectAppVersion(projectRoot?: string): Promise<string> {
   return new Promise<string>((resolve, reject) => {
     try {
-      const projectRoot: string = process.cwd();
+      projectRoot = projectRoot || process.cwd();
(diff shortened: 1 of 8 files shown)


buptkang:
@sergey-akhalkov I know you try to wrap up everything to improve. But I would suggest in the future, for each improvement, you can make a commit so that it would be easy for others to review. Otherwise, it is difficulty for me to read your mind. 😃

buptkang:
LGTM. We can merge this and test it in the INT environment.

sergey-akhalkov:
@buptkang, sure, thanks for the feedback, regarding the improvements list - please see comments from this PR: https://github.com/Microsoft/mobile-center-cli/pull/299

src/commands/codepush/lib/cordova-utils.ts
```diff
@@ -33,6 +33,10 @@ export function isValidOS(platform: string): boolean {
   }
 }
 
+export function isValidPlatform(platform: string): boolean {
+  return platform.toLowerCase() == "cordova";
```
christav: Should be `===`, not `==`

sergey-akhalkov: Got it, thanks for the feedback

PR #301, opened 2017-11-10
Fixed issues with CodePush commands

max-mironov:
In the standalone CodePush CLI, when performing CodePush Patch, users only need to specify deployment name without release label as mandatory parameter.
Also fixed issue with required args for promote command


lumaxis:
@max-mironov Please don't mention internal bug tracker numbers on public repositories as it's of no use and potentially confusing for other users 🙂

max-mironov:
@lumaxis Ops, thanks for pointing! Will update it and take into account in future :)

lumaxis:
Thanks! 🙏

PR #303, opened 2017-11-14
Added logic to switch to legacy service call on codepush release

sergey-akhalkov:
Current implementation has limitation for 1MB upload size, added switch to legacy env until we fix it on new service


package.json
```diff
@@ -111,6 +112,7 @@
     "uuid": "^3.0.1",
     "which": "^1.3.0",
     "wordwrap": "^1.0.0",
-    "xml2js": "^0.4.19"
+    "xml2js": "^0.4.19",
+    "superagent": "^1.7.2"
```
christav: We already have request in the dependency list, why did you drag in yet another http client library?

src/commands/codepush/lib/environment.ts
```diff
@@ -0,0 +1,44 @@
+export interface EnvironmentInfo {
```
christav: These are going to be confusing with the EnvironmentInfo class already declared in the profile system. Wouldn't it be easier to integrate this with the existing profile/environment stuff?

src/commands/codepush/lib/release-command-skeleton.ts
```diff
@@ -81,31 +106,25 @@ export default class CodePushReleaseCommandSkeleton extends AppCommand {
...
-            mandatory: this.mandatory, 
-            noDuplicateReleaseError: this.noDuplicateReleaseError,
-            rollout: this.rollout,
-          },
-          cb)));
-
+      var user = await getUser();
+      const serverUrl = environments(this.environmentName || user.environment).managementEndpoint;
+      
+      await this.releaseStrategy.release(client, this.app, this.deploymentName, updateContentsZipPath, {
```
christav: Shouldn't this be wrapped in an `out.progress` call?

package.json
max-mironov: Make sense, thanks for pointing!

src/commands/codepush/lib/release-command-skeleton.ts
max-mironov: Got it, will do

src/commands/codepush/lib/environment.ts
max-mironov: @christav The main idea was not to mix codepush configuration stuff with other AppCenter CLI configuration.
I mean here that original `EnvironmentInfo ` is different and also not sure if it will be good if we collide codepush with other configs. Also it seems that this solution (with using legacy service) could be not be needed in future and we will simply throw it away later at some point. 
May be we can just rename it e.g. to `CodePushEnvironmentInfo` and keep as it is now? Please let me know what do you think!

christav: I'd be happy with renaming it to `CodePushEnvironmentInfo`.

src/commands/codepush/lib/legacy-codepush-service-client.ts
```diff
@@ -0,0 +1,108 @@
+import * as Request from "request";
```
christav: nitpick: This should be imported as `request`, not `Request`. Upper case is for class constructors, request is a regular function.

src/commands/codepush/lib/release-command-skeleton.ts
```diff
@@ -1,14 +1,29 @@
...
 import * as pfs from "../../../util/misc/promisfied-fs";
 import * as chalk from "chalk";
 import { sign, zip } from "../lib/update-contents-tasks";
 import { isBinaryOrZip } from "../lib/file-utils";
+import { environments } from "../lib/environment";
 import { isValidVersion, isValidRollout, isValidDeployment } from "../lib/validation-utils";
-import * as debug from "debug";
+import { AppCenterCodePushRelease, LegacyCodePushRelease }  from "../lib/release-strategy/index";
+
+const debug = require("debug")("mobile-center-cli:commands:codepush:release-skeleton");
```
christav: Change the debug string to `"appcenter-cli:..."`

src/commands/codepush/lib/legacy-codepush-service-client.ts
christav: Nitpick: Should be imported as `request` not `Request`. Uppercase is for class constructors.

src/commands/codepush/lib/legacy-codepush-service-client.ts
```diff
@@ -0,0 +1,108 @@
...
+import { DefaultApp } from "../../../util/profile/index";
+import * as fs from "fs";
+
+export interface PackageInfo {
+  appVersion?: string;
+  description?: string;
+  isDisabled?: boolean;
+  isMandatory?: boolean;
+  /*generated*/ label?: string;
```
christav: ❓ What does this comment mean?

src/commands/codepush/lib/release-command-skeleton.ts
christav: Please change debug string to `"appcenter-cli:commands:..."`

src/commands/codepush/lib/update-contents-tasks/zip.ts
```diff
@@ -33,20 +34,24 @@ export default function zip(updateContentsPath: string): Promise<string> {
       });
     });
 
-    const packagePath: string = path.join(process.cwd(), generateRandomFilename(15) + ".zip");
-    let zip = new JsZip();
+    var packagePath: string = path.join(process.cwd(), generateRandomFilename(15) + ".zip");
```
christav: Use `let`, not `var`. Unless it can be `const`.

src/commands/codepush/lib/update-contents-tasks/zip.ts
```diff
@@ -2,6 +2,7 @@ import * as fs from "fs";
 import * as pfs from "../../../../util/misc/promisfied-fs";
 import * as path from "path";
 import * as JsZip from "jszip";
+var yazl = require("yazl");
```
christav: No typings? 

If you have to use require, use `const` for the variable, not `var`.

And what is the compelling value of yazl over jszip that you need to bring in a second package with repeated functionality?

src/commands/codepush/lib/environment.ts
sergey-akhalkov: fixed

src/commands/codepush/lib/legacy-codepush-service-client.ts
sergey-akhalkov: fixed

src/commands/codepush/lib/legacy-codepush-service-client.ts
sergey-akhalkov: removed

src/commands/codepush/lib/release-command-skeleton.ts
sergey-akhalkov: fixed

src/commands/codepush/lib/update-contents-tasks/zip.ts
sergey-akhalkov: There is no original typings, so I've added custom typings from the old code-push-cli https://github.com/Microsoft/code-push/blob/master/definitions/yazl.d.ts

Regarding the repeated functionality - this hacky solution is needed to consistent hashes calculation both on codepush clients SDKs and on the old codepush service. Old codepush service uses yazl lib to work with zip archives and for some reason the hashes are calculated wrong for the zip packages produced with jszip lib, but works correctly with yazl.

So I'm absolutely agreed that the yazl contains repeated functionality but unfortunately we need it until we not figured out what the difference between yazl and jszip in terms of hashes calculation. Another words it's a temporary solution that hopefully will be removed soon.

src/commands/codepush/lib/update-contents-tasks/zip.ts
sergey-akhalkov: my bad, fixed, thanks

src/commands/codepush/lib/update-contents-tasks/zip.ts
christav: Fair enough, that's a reasonable explanation. Thanks!

Although - how do we know that yazl wasn't doing it wrong rather than jszip? 😄 As long as it works, I guess it's ok.

PR #311, opened 2017-11-17
[codepush] fix for cordova-release command

sergey-akhalkov:
Update/Release for Cordova has another structure than React updates:
1. Cordova update should contain www/ root folder inside the archive
2. React update should contain CodePush/ root folder inside the archive

### DIFF
--- a/src/commands/codepush/release-react.ts
+++ b/src/commands/codepush/release-react.ts
@@ -3,6 +3,7 @@ import CodePushReleaseCommandSkeleton from "./lib/release-command-skeleton"
 import { AppCenterClient, models, clientRequest } from "../../util/apis";
 import { out } from "../../util/interaction";
 import { inspect } from "util";
+import * as fs from "fs";
 import * as pfs from "../../util/misc/promisfied-fs";
 import * as chalk from "chalk";
 import * as path from "path";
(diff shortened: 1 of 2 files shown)


src/commands/codepush/release-react.ts
```diff
@@ -85,6 +85,10 @@ export default class CodePushReleaseReactCommand extends CodePushReleaseCommandS
     this.platform = appInfo.platform.toLowerCase();
 
     this.updateContentsPath = this.outputDir || await pfs.mkTempDir("code-push");
+    
+    // we have to add "CodePush" root forlder to make update contents file structure 
+    // to be compatibale with React Native client SDK
```
christav: Spelling - `compatible`

src/commands/codepush/release-react.ts
```diff
@@ -85,6 +85,10 @@ export default class CodePushReleaseReactCommand extends CodePushReleaseCommandS
...
+    await pfs.mkdir(path.join(this.updateContentsPath, "CodePush"));
```
christav: Doesn't need to be async.

src/commands/codepush/release-react.ts
sergey-akhalkov: fixed

src/commands/codepush/release-react.ts
sergey-akhalkov: fixed

PR #316, opened 2017-11-22
Test cloud error message improvements

owenniblock:
### Motivation
Users have been sending just the environment message, so to help support we should explicitly ask for both the environment message and the error message before hand.

### Changes

- Slight alteration to blanket error message to ask for both env message and error message.
- Better messages for test-cloud.exe error codes.

### DIFF
--- a/src/commands/test/lib/run-tests-command.ts
+++ b/src/commands/test/lib/run-tests-command.ts
@@ -154,7 +154,7 @@ export abstract class RunTestsCommand extends AppCommand {
       let message : string = null;
       let profile = getUser();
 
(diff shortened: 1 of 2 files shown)


src/commands/test/lib/run-tests-command.ts
```diff
@@ -154,7 +154,7 @@ export abstract class RunTestsCommand extends AppCommand {
       let message : string = null;
       let profile = getUser();
 
-      let helpMessage = `Further error details: For help, please send the following information to us by going to https://mobile.azure.com/apps and starting a new conversation (using the icon in the bottom right corner of the screen)${os.EOL}
+      let helpMessage = `Further error details: For help, please send both the reported error above and the following environment information to us by going to https://mobile.azure.com/apps and starting a new conversation (using the icon in the bottom right corner of the screen)${os.EOL}
```
christav: Wrong url in help message.

Issue #297, opened 2017-11-09
The version option is not discoverable


krukow:
The `mobile-center` CLI supports a `--version` or `-v` option. However, this is not discoverable from the help output:

```sh
➜  code mobile-center help

Mobile Center helps you build, test, distribute, and monitor mobile apps.

Usage: mobile-center <command>

Commands:
    analytics                      View events, audience info, sessions, and
                                   other anayltics for apps
    apps                           View and manage apps
    build                          Start builds, get their status, and
                                   download artifacts
    codepush                       View and manage CodePush deployments and
                                   releases
    crashes                        Upload symbols for better crash reports
    distribute                     Send builds to testers and manage
                                   distribution groups
    orgs                           Manage organizations
    profile                        Manage your profile
    telemetry                      Manage telemetry preferences
    test                           Start test runs and get their status
    tokens                         Manage API tokens
    help                           Get help using mobile-center commands
    login                          Log in
    logout                         Log out
    setup-autocomplete             Setup tab completion for your shell
```

jmoody:
While we are at it:

```
# Expected
$ mobile-center --version
0.12.0

# Found
$ mobile-center --version
mobile-center version 0.12.0
```

Why the additional text?

amchew:
Hey @krukow and @jmoody, this feedback is good. We're planning to edit the help text to include: 1. the version and 2. a list of switches and their descriptions that works with every CLI command, and are working on it next sprint. Thanks again for the useful feedback.

christav:
As of #347 the version number is now printed when you look at the all-up `appcenter` help.

PR #318, opened 2017-11-23
Print version info command in help output

KevinTCoughlin:
👋

Attempts to address #297 by printing version command similar to azure-cli:

```shell
$ node .\dist\index.js -h

Visual Studio App Center helps you build, test, distribute, and monitor mobile apps.

For version info, use 'appcenter --version'

Usage: appcenter <command>

Commands:
    analytics                      View events, audience info, sessions, and other anayltics for apps
    apps                           View and manage apps
    build                          Start builds, get their status, and download artifacts
    codepush                       View and manage CodePush deployments and releases
    crashes                        Upload symbols for better crash reports
    distribute                     Send builds to testers and manage distribution groups
    orgs                           Manage organizations
    profile                        Manage your profile
    telemetry                      Manage telemetry preferences
    test                           Start test runs and get their status
    tokens                         Manage API tokens
    help                           Get help using appcenter commands
    login                          Log in
    logout                         Log out
    setup-autocomplete             Setup tab completion for your shell
```

Thoughts? Admittedly is a bit odd to me printing before `Usage`.

### DIFF
--- a/src/util/commandline/category-command.ts
+++ b/src/util/commandline/category-command.ts
@@ -26,6 +26,8 @@ export class CategoryCommand extends Command {
     out.help();
     out.help(this.categoryHelp());
     out.help();
+    out.help("For version info, use 'appcenter --version'");
+    out.help();
     const command = "<command>";
     const commandTemplate = [scriptName].concat(this.command, [command]).join(" ");
     out.help(`Usage: ${chalk.bold(commandTemplate)}`);


christav:
This is an interesting one, actually. I'm actually thinking of fixing this by adding a "global switches" section to the bottom of all the help pages which will include --version, --output, --debug, etc. That way we can manage the list automatically rather than having to manually edit the help.

Thoughts?

christav:
Another thought - we can just print the version number itself as part of the help message.

KevinTCoughlin:
👋 @christav I think both are great additions and would satisfy the initial issue :). I've seen both quite a bit in the wild myself.

christav:
We just print the version directly in the help now, so no need for this change.

PR #320, opened 2017-11-23
Typo

kant:


christav:
Thanks for the PR, but this isn't the right fix. That section of the readme is generated from the command help automatically (see the scripts/get-commands.js file), so editing the readme will get overwritten in the next release.

The real fix is to fix the spelling error in the help text for the command.

PR #321, opened 2017-11-23
[BugFix] CodePush: Failed to make a success release

buptkang:
This PR fixes the issue to make a codepush release. Tested in both org/non org apps.

### DIFF
--- a/src/commands/codepush/lib/legacy-codepush-service-client.ts
+++ b/src/commands/codepush/lib/legacy-codepush-service-client.ts
@@ -40,7 +40,7 @@ export default class LegacyCodePushServiceClient {
   }   
 
   public release(deploymentName: string, filePath: string, updateMetadata: PackageInfo): Promise<void> {   
-    const appName = this._app.appName;
+    const appName = this._app.ownerName + '/' + this._app.appName;
     this._debug(`Releasing update via old service to ${this.appNameParam(appName)} app ${deploymentName} deployment`);
 
     return new Promise<void>((resolve, reject) => {


max-mironov:
Hey @buptkang, looks good for me, just a nitpick - we probably can move appName to properties, will look better and be more reusable if needed. I've created PR for this in your repo https://github.com/buptkang/mobile-center-cli/pull/1 please take a look and merge if you are ok with this!

buptkang:
Thanks @max-mironov , the change looks good with testing passed. But next time, we should make sure to test after each code merge. Right now, the codepush release-react did not work because we did not test intensively. This affects users to use the codepush in the app center now.

src/commands/codepush/lib/legacy-codepush-service-client.ts
```diff
@@ -39,8 +39,12 @@ export default class LegacyCodePushServiceClient {
     return this._accessKey;
   }   
 
+  private get appIdentifer(): string {
```
sergey-akhalkov: Do we really need this? `_app` instance of `DefaultApp` class already has `identifier` property, for example
https://github.com/Microsoft/appcenter-cli/blob/552e0f7d17316aaa1fc3e7fed8908c7c6d4d65b8/src/commands/apps/delete.ts#L15

max-mironov: Good catch, missed this, thanks for pointing!
@buptkang agreed with what you said above! Hope this PR and other CodePush related PR's will be merged ASAP.

christav: Yeah, I'd prefer if you removed this and just used this._app.identifier instead. Also, since I can't make a comments on the individual lines for some reason if they're not part of the changes, some additional comments before merging this:

Line 2 - "remove `/index` from the import path, it's automatic. In general, import the directory as a package and leave it up to the package author to expose things.

Line 26: You could use parameter properties on your constructor and save a bunch of variables - for example do this instead:
```
  constructor(private accessKey: string, private serverUrl: string, private app: DefaultApp, private debug: Function, private customHeaders?: Headers)
```

And then remove all the `this._accessKey = accessKey` lines, and all the `_` properties.

Line 26: DO NOT PASS IN THE DEBUG FUNCTION! The whole point of debug is we know what line in which file it is beign printed from, this breaks that. Create your own debug function by requiring the debug module the same way every other file does it.

buptkang:
@christav @sergey-akhalkov this PR's main purpose is to fix the issue and unblock users to use the service. Currently, there are at least three users from intercom complaining the issue.  After this PR merged, @max-mironov can continue to work on the existing code to improve the code quality.

PR #325, opened 2017-11-27
Make @types/* packages dev dependencies

yuya-takeyama:
IMHO, `@types/*` packages always should be `devDependencies`.

I added this package to `devDependencies`.
When I run `yarn install`, it installed these `@types*` packages needlessly.


christav:
I had not really thought about this, but you make an excellent point and this is a good idea.

PR #326, opened 2017-11-27
[CodePush] fixed the debug function passing and private properties code

buptkang:
This PR fixes the issue from the previous PR #321 that @christav pointed out.

This PR fixed:

1. private constructor properties mentioned in [here](https://github.com/Microsoft/appcenter-cli/pull/321#discussion_r153288656). 
2. remove debug function to pass through [here](https://github.com/Microsoft/appcenter-cli/pull/321#discussion_r153288656).

@christav for the comment to not passing DefaultApp. Since codepush release command points to legacy one instead of the new one, this requirement makes us to think to pass the DefaultApp from higher level command to the lower level skeleton code. In other word, we do the code injection in purpose.  @max-mironov this is what I think from your written code.

### DIFF
--- a/src/commands/codepush/lib/legacy-codepush-service-client.ts
+++ b/src/commands/codepush/lib/legacy-codepush-service-client.ts
@@ -17,35 +17,17 @@ export type Headers = { [headerName: string]: string };
 export default class LegacyCodePushServiceClient {
   private static API_VERSION: number = 2;
   
-  private _accessKey: string;
-  private _serverUrl: string;
-  private _customHeaders: Headers;
-  private _app: DefaultApp;
-  private _debug: Function;
-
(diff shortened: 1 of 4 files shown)


max-mironov:
@buptkang thanks for taking care!
@christav  Thanks for pointing out, that really make sense, as for passing in debug - we probably should fix it in other places too e.g. https://github.com/Microsoft/appcenter-cli/blob/0cc7768bd954c17d705431374a63163d8ce20521/src/commands/crashes/lib/azure-blob-upload-helper.ts#L8 , https://github.com/Microsoft/appcenter-cli/blob/552e0f7d17316aaa1fc3e7fed8908c7c6d4d65b8/src/commands/crashes/lib/symbols-uploading-helper.ts#L8 and probably some other places.

Issue #328, opened 2017-11-28
React Native - Deleting output folder after release


renatoagds:
Hi!

Today I migrated from `code-push-cli` to `appcenter-cli`, but I'm having a problem, I use the generated bundle and sourcemap to upload them for `sentry`:

https://docs.sentry.io/clients/react-native/codepush/

But, in the final of the release process the `appcenter-cli` delete the generated folder:

https://github.com/Microsoft/appcenter-cli/blob/master/src/commands/codepush/release-react.ts#L153

I want to propose a flag to avoid delete the output dir, so I could keep the files after a release.

ruslan-bikkinin:
Hi @renatoagds and thanks for submitting the issue. I've been fixed it here: #329 Do you have a chance to take a look at it and check whether it works for you or you have any more issues with that?

henrikra:
Please fix this :D It does not make sense that we are defining output directory and then it is deleted at the end :O We want also send our sourcemaps to Bugsnag

renatoagds:
@ruslan-bikkinin hey! thanks for the quick response! I'll test today it and send the feedback. 👍

renatoagds:
@ruslan-bikkinin tested your branch and worked like a charm, waiting for a new release. 🚀

sergey-akhalkov:
Hi @christav, could you please merge this PR in addition to this one https://github.com/Microsoft/appcenter-cli/pull/331 and probably make a new CLI release?

christav:
I merged the PR, I'll do a new release on Monday.

christav:
Release 1.0.6 is out with this fix. I'm closing this, if 1.0.6 doesn't fix the issue please reopen this.

PR #329, opened 2017-11-29
Fix final deletion of output directory after codepush release-react

ruslan-bikkinin:
This PR including fix for #328


henrikra:
Hmm try to do release with your change and after that make another. I usually get error then. I guess this has something to do that the output folder is there and it is not empty or something

ruslan-bikkinin:
@henrikra got it. This is regression after moving codepush to appcenter, I'll fix it in the next PR. For now you should create output directory manually like this: `mkdir -p /your/out/dir`.

PR #330, opened 2017-11-29
Updates the UITest template zip location to point to 1.1

owenniblock:
Updates generate command to use new templates.

### DIFF
--- a/src/commands/test/generate/uitest.ts
+++ b/src/commands/test/generate/uitest.ts
@@ -14,6 +14,6 @@ export default class GenerateUITestCommand extends GenerateCommand {
     super(args);
   }
 
-  protected zipPathAndroid = "UITest/Android-1.0.zip";
-  protected zipPathiOS = "UITest/iOS-1.0.zip";
+  protected zipPathAndroid = "UITest/Android-1.1.zip";
+  protected zipPathiOS = "UITest/iOS-1.1.zip";
 }


christav:
@krukow Thanks, I'll merge and release tomorrow - well, later today :-) - morning.

owenniblock:
I've confirmed that `ts-node src/index.ts test generate uitest --platform android --output-path ~/tmp-gen` pulls down the latest .zip - I'm also not entirely sure how this is used by the launch test. Should we find that out before this goes live?

krukow:
Yes, this goes through the normal QA process

owenniblock:
Don't merge this yet :) Explanation below.

owenniblock:
Talking to Tomasz I've now got a better understanding of the build process behind the launch test.

There is a reference to the project name in one of the build TS files. I will detail this on the VSTS ticket but the upshot is that before we release this to the CLI we need to:
1. Update this reference in the script.
2. Do some magic to migrate users who already have this script.

We'll have to get all this in place before we release this change to the cli because the script currently takes the _latest_ version of the appcenter-cli.

I would suggest we patch Android-1.0.zip and iOS-1.0.zip with the latest UITest to fix the broken package and then plan this out given the new knowledge.

krukow:
@owenniblock thanks for the update

krukow:
@owenniblock @christav publishing this is safe - please merge and publish

owenniblock:
Ok, I've confirmed with the build team this won't affect customers. So we're good to go with the original plan - sorry for the delay but I wanted to be cautious after talking to Tomasz :)

PR #331, opened 2017-11-30
Fix validation of `--target-binary-version`

yuya-takeyama:
Currently, I can't specify `--target-binary-version` using semver range.

```
$ appcenter codepush release-react -a MyOrg/MyApp --deployment-name Staging --target-binary-version '1.2.x'
Error: Invalid binary version(s) for a release.
```

The document says strings `1.2.x` should be allowed.
https://docs.microsoft.com/en-us/appcenter/distribution/codepush/cli#target-binary-version-parameter

If the version is like `1.2.0`, it runs successfully.

I figured out the cause of the error and fixed it.
It's because of the usage of `semver` npm package.

```js
var semver = require('semver');

console.dir(semver.valid('1.2.0'));
// => '1.2.0

console.dir(semver.valid('1.2.x'));
// => null

console.dir(semver.validRange('1.2.0'));
// => '1.2.0

console.dir(semver.validRange('1.2.x'));
// => '>=1.2.0 <1.3.0'

console.dir(semver.validRange('1.2.invalid'));
// => null
```

### DIFF
--- a/src/commands/codepush/lib/validation-utils.ts
+++ b/src/commands/codepush/lib/validation-utils.ts
@@ -4,12 +4,12 @@ import { AppCenterClient, models, clientRequest } from "../../../util/apis";
 import { DefaultApp } from "../../../util/profile";
 
 // Check if the given string is a semver-compliant version number (e.g. '1.2.3')
+// (missing minor/patch values will be added on server side to pass semver.satisfies check)
 export function isValidVersion(version: string): boolean {
-  return !!semver.valid(version);
+  return !!semver.valid(version) || /^\d+\.\d+$/.test(version) || /^\d+$/.test(version);
 }
 
 // Allow plain integer versions (as well as '1.0' values) for now, e.g. '1' is valid here and we assume that it is equal to '1.0.0'. 
-// (missing minor/patch values will be added on server side to pass semver.satisfies check)
 export function isValidRange(semverRange: string): boolean {
   return !!semver.validRange(semverRange);
 }
--- a/src/commands/codepush/lib/validation-utils.ts
+++ b/src/commands/codepush/lib/validation-utils.ts
@@ -3,9 +3,15 @@ import * as semver from "semver";
 import { AppCenterClient, models, clientRequest } from "../../../util/apis";
 import { DefaultApp } from "../../../util/profile";
 
+// Check if the given string is a semver-compliant version number (e.g. '1.2.3')
+// Ranges (e.g. '1.2.x', '>= 1.2.0' etc) are not allowed
+export function isValidVersion(version: string): boolean {
+  return !!semver.valid(version);
+}
+
 // Allow plain integer versions (as well as '1.0' values) for now, e.g. '1' is valid here and we assume that it is equal to '1.0.0'. 
 // (missing minor/patch values will be added on server side to pass semver.satisfies check)
-export function isValidVersion(semverRange: string): boolean {
+export function isValidRange(semverRange: string): boolean {
   return !!semver.validRange(semverRange) || /^\d+\.\d+$/.test(semverRange) || /^\d+$/.test(semverRange);
 }
 
--- a/src/commands/codepush/lib/validation-utils.ts
+++ b/src/commands/codepush/lib/validation-utils.ts
@@ -6,7 +6,7 @@ import { DefaultApp } from "../../../util/profile";
 // Allow plain integer versions (as well as '1.0' values) for now, e.g. '1' is valid here and we assume that it is equal to '1.0.0'. 
 // (missing minor/patch values will be added on server side to pass semver.satisfies check)
 export function isValidVersion(semverRange: string): boolean {
-  return !!semver.valid(semverRange) || /^\d+\.\d+$/.test(semverRange) || /^\d+$/.test(semverRange);
+  return !!semver.validRange(semverRange) || /^\d+\.\d+$/.test(semverRange) || /^\d+$/.test(semverRange);
 }
 
 export function isValidRollout(rollout: number): boolean {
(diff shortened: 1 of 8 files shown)


max-mironov:
Hey @yuya-takeyama thanks so much for contribution!

You are correct in your assumptions however I believe we need to consider several more things here:

a) I think we can rename (or add new func `isValidRange`, see other comments below) `isValidVersion` to `isValidRange` function

b) We can remove for `isValidRange` the following:
```
/^\d+\.\d+$/.test(semverRange) || /^\d+$/.test(semverRange)
```
because e.g. `semver.validRange('1')` will also handle this cases 

c) But we need to check these conditions when extracting appversion from plist/gradle file. 
So this way we need one more utility function e.g. `isValidVersion` in this case in `validation-utils`. So it will look exactly like it is now. 

Will use this function in all occurrences for https://github.com/Microsoft/appcenter-cli/blob/master/src/commands/codepush/lib/react-native-utils.ts#L22 e.g. like this one: 
https://github.com/Microsoft/appcenter-cli/blob/master/src/commands/codepush/lib/react-native-utils.ts#L74

Please let me know what do you think about it and if you have any questions/need any clarifications.

We appreciate if you push this changes by yourself but just let us know if you need any assistance from our side and we can apply additional changes on our side.

yuya-takeyama:
@max-mironov 
Totally agreed!
Let me do that by myself.

yuya-takeyama:
@max-mironov 
Addressed!

BTW, is this condition correct?
Why not just `else`?
https://github.com/yuya-takeyama/appcenter-cli/blob/d3b435344a03fb7732e737245613723bc2325aac/src/commands/codepush/lib/react-native-utils.ts#L132-L137

src/commands/codepush/lib/validation-utils.ts
```diff
@@ -3,10 +3,15 @@ import * as semver from "semver";
 import { AppCenterClient, models, clientRequest } from "../../../util/apis";
 import { DefaultApp } from "../../../util/profile";
 
+// Check if the given string is a semver-compliant version number (e.g. '1.2.3')
+export function isValidVersion(version: string): boolean {
+  return !!semver.valid(version);
```
max-mironov: We should also validate for `/^\d+\.\d+$/.test(semverRange) || /^\d+$/.test(semverRange)` due to we also allow integer numbers lile 1.0, 1 etc...
also we need to move comment from `/^\d+\.\d+$/.test(semverRange) || /^\d+$/.test(semverRange)` about plain integer version number usage here

max-mironov:
@yuya-takeyama for the condition - seems that error message from `else`
```
The "android.defaultConfig.versionName" property in the "${buildGradlePath}" file needs to specify a valid semver string, containing both a major and minor version (e.g. 1.3.2, 1.1).`
``` 
is not relevant now.
I believe we should remove it at all, as currently plain integer numbers are allowed.

src/commands/codepush/lib/validation-utils.ts
yuya-takeyama: Thanks, addressed!
https://github.com/Microsoft/appcenter-cli/pull/331/commits/b76a0af672a851edf70f8ad9e1b7891cf6be7b1f

yuya-takeyama:
@max-mironov 
So now, should it be like below?:

```ts
        if (isValidVersion(appVersion)) {
          if (!/^\d.*/.test(appVersion))
						// The versionName property isn't a valid semver string,
						// but it starts with a number, and therefore, it can't
						// be a valid Gradle property reference.
						throw new Error(`The "android.defaultConfig.versionName" property in the "${buildGradlePath}" file needs to specify a valid semver string, containing both a major and minor version (e.g. 1.3.2, 1.1).`);
          }

          // The versionName property is a valid semver string,
          // so we can safely use that and move on.
          out.text(`Using the target binary version value "${appVersion}" from "${buildGradlePath}".\n`);
          return appVersion;
        }
```

Sorry, I still didn't get the expected return-value of the method (`getReactNativeProjectAppVersion`).

Or is it better to have another specific-method for checking semver strictly?

```ts
export function isStrictlyValidVersion(version: string): boolean {
  return !!semver.valid(version);
}

export function isValidVersion(version: string): boolean {
  return !!semver.valid(version) || /^\d+\.\d+$/.test(version) || /^\d+$/.test(version);
}
```

max-mironov:
@yuya-takeyama well, I'd rather to remove this check at all as we can pass integer number, so it would be like this:
```
...
let appVersion: string = versionName.replace(/"/g, "").trim();

        if (isValidVersion(appVersion)) {
          // The versionName property is a valid semver string,
          // so we can safely use that and move on.
          out.text(`Using the target binary version value "${appVersion}" from "${buildGradlePath}".\n`);
          return appVersion;
        } 
        // The version property isn't a valid semver string
        // so we assume it is a reference to a property variable.
        const propertyName = appVersion.replace("project.", "");
        const propertiesFileName = "gradle.properties";
...
```

yuya-takeyama:
@max-mironov 
From the error message:

>a valid semver string, containing both a major and minor version (e.g. 1.3.2, 1.1).

It seems like versions lacking `minor version` (like `1`) should not be allowed.
But `isValidVersion` allows it.
Is that correct?

max-mironov:
@yuya-takeyama 
It was previously but later we've added support for plain integer numbers. Please see this PR for details: https://github.com/Microsoft/code-push/pull/429
So for now it is ok to use it and seems that error message is outdated and not relevant now.

yuya-takeyama:
@max-mironov 
Thanks, I got your intention now!
Addressed: https://github.com/Microsoft/appcenter-cli/pull/331/commits/20b2da21c54e775f1840559704e833ef6929f4c6

max-mironov:
@yuya-takeyama 
Thank you so much for your input and attention to details, appreciate it!

christav:
@max-mironov Are you ok with the state of this PR now? Should it be merged or should I hold off?

sergey-akhalkov:
Hey @christav, both me and Max think that we are good with this PR, please feel free to merge.

PR #339, opened 2017-12-08
Update login text

jngli:


src/commands/login.ts
```diff
@@ -103,7 +103,7 @@ export default class LoginCommand extends Command {
   private async doInteractiveLogin(): Promise<TokenValueType> {
     const loginUrl = environments(this.environmentName).loginEndpoint + "?" + qs.stringify({ hostname: os.hostname()});
 
-    out.text(`Opening browser to log in. If the browser does not open, please go to ${loginUrl}, log in, and enter the code returned.`);
+    out.text(`Opening your browser... \n? Visit ${loginUrl} and enter the code:`);
```
ruslan-bikkinin: I believe there should be [os.EOL](https://nodejs.org/api/os.html#os_os_eol) used instead of plain `\n`.

PR #342, opened 2017-12-11
Feature/validate token on create client

owenniblock:
### Motivation
Currently if the wrong --token is sent - the system doesn't flag this up until the first API call. This PR is to check an API call sooner and return a helpful error message.

### Disclaimer
I'm relatively new to TypeScript - please let me know if I've done anything incorrect here. I've had to:
1. Convert a load of stuff to async so we can access the API.
2. Introduce some constructor injection to pass through a stubbed `tokenValidator` so tests still pass.

I'm creating this PR early for discussion so please let me know if anything needs to change for acceptance.

### DIFF
--- a/src/commands/login.ts
+++ b/src/commands/login.ts
@@ -89,7 +89,7 @@ export default class LoginCommand extends Command {
     }
 
     const endpoint = environments(this.environmentName).endpoint;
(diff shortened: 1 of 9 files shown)


jngli:
Closed this one according to the discussion in #344

PR #344, opened 2017-12-12
Fix/bad error on 401

owenniblock:
### Motivation
Currently 401 errors are badly reported by the system.

### Solution
This PoC throws a better error when the first operation on the run test process fails and then reports that error to the user.

If we decide to go with this method - we need to:
1. Refactor the error catching code used whenever the API service is called (so that a consistent approach to 401s can be applied.
2. Update all areas of code where the API service call is used to use this new function.


owenniblock:
No objection at all. I've refactored the error handling code a little so you can see where I'm headed with that. If it looks OK I'll apply it to the other places where the API is called so I can check for edge cases.

PR #348, opened 2017-12-15
adding `common` property decorator to switch options and use it to disable better example and options

Huan9:
screen shot 2017-12-14 at 3 11 11 pm

### DIFF
--- a/src/util/commandline/command.ts
+++ b/src/util/commandline/command.ts
@@ -1,6 +1,6 @@
 // Base class for all command handlers
 import * as Result from "./command-result";
-import { shortName, longName, help, hasArg, getOptionsDescription, getPositionalOptionsDescription } from "./option-decorators";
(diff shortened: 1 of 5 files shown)


christav:
I'd like to see a separation in the options list between the command specific and the common options, so ti looks something like:
```
Options:
  -b|--branch <arg>   Branch name for status check
  -a|--app <arg>        Specify app

Common Options (works on all commands):
     --disable-telemetry   
  -v|--version                   Display appcenter version
```
etc.

Huan9:
The new after and before base on the suggestion

screen shot 2017-12-15 at 12 08 12 pm

christav:
Code looks good! @amchew does this new output look good to you?

amchew:
@Huan9 @christav , yes this looks good to me. Thanks for the quick work Huan and Chris 👍 !

PR #352, opened 2018-01-03
UITest: adds ability to specify extra parameters to prepare uitest

owenniblock:
### Motivation
We added the missing parameters to the main `run uitest` method and this makes the equivalent changes when the user calls `prepare` followed by `run manifest`.

We need this to support `--include-category` and `--exclude-category` in the VSTS Task among other client requests.

### Notes
I moved `fixArrayParameter` out to the shared `Command` but please let me know if you'd rather see this moved out somewhere else.


owenniblock:
Great, thanks @lumaxis - I don't have access to `merge` could someone please do the honours? 🙂

PR #358, opened 2018-01-12
Feature/profile reads env vars for local

sapieneptus:
### Motivation

Previously there was a `testCloudLocalDev` environment option with some hardcoded endpoints. I've generalized this environment configuration to read from the `process.env` and renamed it to just `local`. 

The intention is that now a developer could easily specify where `appcenter-cli` should search for endpoints e.g. on their local machine. This is convenient when, e.g., we need to run the portal/api on alternative ports/hostnames. 

Example:
```shell
APPCENTER_PORTAL_ENDPOINT=<local location> APPCENTER_ENDPOINT=<local location> \
appcenter login --env local --debug -u <name> -p <pw>
```

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,6 +1,6 @@
 {
   "name": "appcenter-cli",
-  "version": "1.0.6",
+  "version": "1.0.7",
   "lockfileVersion": 1,
   "requires": true,
   "dependencies": {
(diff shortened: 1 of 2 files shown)


sapieneptus:
CC: @christav @john7doe

christav:
@sapieneptus Note to self & you: Run the test suite before merging checkins. This change broke a bunch of the tests. I'll fix it for now, but please be sure to run the tests locally in the future before submitting PR's.

sapieneptus:
Wouldn’t have expected breakages - duly noted. Thanks for letting me know.

sapieneptus:
@christav would you mind tagging me in the fix commit(s)?

PR #363, opened 2018-01-16
Replaced generated clients with new one

georgii-ivanov:
- Removed previous clients;
- Created lib/* directory location for placing local libraries;
- Temporary added clients as local package to package.json;
- Fixed all commands definitions for new clients API support;
- Updated version of ms-rest package.

Note: The generated clients needs to be moved to new public repo.


christav:
Closing this until we reprioritize swagger work.

PR #368, opened 2018-01-23
Updated to latest autorest.

christav:
Updated code to call new generated client code.
Update build system to use latest autorest.
Updated build to run everything from npm, removed autorest from
gulpfile.


buptkang:
👍

PR #371, opened 2018-01-25
Distribution Release API 

buptkang:
Since the backend file upload service has been updated, this PR updates the cli side to synchronize with the backend side.

### DIFF
--- a/package.json
+++ b/package.json
@@ -39,6 +39,7 @@
     "url": "https://github.com/Microsoft/appcenter-cli/issues"
   },
   "devDependencies": {
+    "@types/progress": "^2.0.1",
     "@types/async": "^2.0.42",
     "@types/chai": "^4.0.4",
     "@types/chai-as-promised": "^7.1.0",
@@ -82,6 +83,7 @@
     "typescript": "^2.5.2"
   },
   "dependencies": {
+    "appcenter-file-upload-client": "0.0.5",
     "azure-storage": "^2.1.0",
     "bplist": "0.0.4",
     "chalk": "^1.1.3",
@@ -105,6 +107,7 @@
     "opener": "^1.4.3",
     "p-limit": "^1.0.0",
     "plist": "^2.1.0",
+    "progress": "^2.0.0",
     "properties": "^1.2.1",
     "qs": "^6.4.0",
     "request": "^2.76.0",
--- a/src/commands/distribute/release.ts
+++ b/src/commands/distribute/release.ts
@@ -1,4 +1,4 @@
-import { AppCommand, CommandResult, ErrorCodes, failure, hasArg, help, longName, required, shortName, success } from "../../util/commandline";
+import { AppCommand, CommandResult, ErrorCodes, failure, hasArg, help, longName, required, shortName, success, failed } from "../../util/commandline";
 import { AppCenterClient, models, clientRequest, ClientResponse } from "../../util/apis";
 import { out } from "../../util/interaction";
 import { inspect } from "util";
@@ -7,6 +7,12 @@ import * as Request from "request";
 import * as Path from "path";
 import * as Pfs from "../../util/misc/promisfied-fs";
 import { DefaultApp } from "../../util/profile";
+import FileUploadClient, { IFileUploadClientSettings, IProgress, IUploadResults, MessageLevel } from "appcenter-file-upload-client";
+import { UploadProgress } from "../../util/apis/generated/models/index";
+import { fail } from "assert";
+import { setTimeout } from "timers";
+
+var ProgressBar = require('progress');
 
 const debug = require("debug")("appcenter-cli:commands:distribute:release");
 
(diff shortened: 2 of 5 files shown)


package.json
```diff
@@ -82,6 +82,8 @@
     "typescript": "^2.5.2"
   },
   "dependencies": {
+    "@types/progress": "^2.0.1",
```
christav: @types should be a devDependency, not a dependency.

src/commands/distribute/release.ts
```diff
@@ -7,6 +7,12 @@ import * as Request from "request";
 import * as Path from "path";
 import * as Pfs from "../../util/misc/promisfied-fs";
 import { DefaultApp } from "../../util/profile";
+import FileUploadClient, { IFileUploadClientSettings, IProgress, IUploadResults, MessageLevel } from "appcenter-file-upload-client";
```
christav: What is this syntax with one name outside the braces and others inside? What does this do?

src/commands/distribute/release.ts
```diff
@@ -39,50 +45,46 @@ export default class ReleaseBinaryCommand extends AppCommand {
   public releaseNotesFile: string;
 
   public async run(client: AppCenterClient): Promise<CommandResult> {
+
+    const bar = new ProgressBar(' Uploading a new release [:bar] :percent', {
```
christav: How does this progress bar integrate with the `--quiet` switch? Looks like it doesn't. You shouldn't display the progress bar if `--quiet` is given. Look at the stuff in the interaction directory (like out.text, out.progress, etc.) to see the patterns used to handle this.

package.json
buptkang: Oh, good point.

src/commands/distribute/release.ts
buptkang: FileUploadClient is the default export class, whereas the others are not. Here it is its definition: https://github.com/Microsoft/appcenter-file-upload-client/blob/master/src/index.ts

src/commands/distribute/release.ts
```diff
@@ -144,83 +120,34 @@ export default class ReleaseBinaryCommand extends AppCommand {
         throw statusCode;
       }
     } catch (error) {
-      if (error === 404) {
-        throw failure(ErrorCodes.InvalidParameter, `distribution group ${this.distributionGroup} was not found`);
-      } else {
-        debug(`Failed to get users of distribution group ${this.distributionGroup}, returning null - ${inspect(error)}`);
-        return null;
-      }
+      throw failure(ErrorCodes.InvalidParameter, `distribution group ${this.distributionGroup} was not found`);
```
izikl: So if the server returns 500 we show a message "distribution group <name> was not found"? That will be incorrect and misleading. 
We should show a generic error message with a context ID (If context ID is shown when running with `--debug` in `client.distributionGroups.listUsers` then having a context ID printed here is just nice to have).

src/commands/distribute/release.ts
```diff
@@ -144,83 +120,34 @@ export default class ReleaseBinaryCommand extends AppCommand {
...
     try {
-      abortReleaseUploadRequestResponse = await out.progress("Aborting release upload...",
-        clientRequest<models.ReleaseUploadEndResponse>((cb) => client.releaseUploads.complete(uploadId, app.ownerName, app.appName, "aborted", cb)));
+      uploadProgressResponse = await clientRequest<models.UploadProgress>((cb) => client.releaseUploads.get(assetId, 
+        app.ownerName, 
+        app.appName, 
+        cb));
     } catch (error) {
-      throw new Error(`HTTP ${abortReleaseUploadRequestResponse.response.statusCode} - ${abortReleaseUploadRequestResponse.response.statusMessage}`);
+      throw failure(ErrorCodes.Exception, `Processing the uploaded app file failed: ${error}.`);
```
izikl: I would change the error to "Failed to process the package." and again separate 500s from 400s. On 500s the the expectation from the user is ping us in intercom, on 400s we probably want them to change something in the package.

package.json
```diff
@@ -82,6 +82,8 @@
...
+    "appcenter-file-upload-client": "^0.0.6",
```
sergey-akhalkov: Hi @buptkang, please change package version to `0.0.5`, I believe that `>= 0.0.6` is not going to work on Windows due to incorrect superagent dependency. `package.json` of  `appcenter-file-upload-client` should be updated to use the stable version of superagent.

Please let me know if you have any questions.

buptkang: Thanks for this information.

buptkang: Switched back to the previous version.

src/commands/distribute/release.ts
buptkang: that is a really good point, have abstracted the method.

src/commands/distribute/release.ts
```diff
@@ -39,50 +45,46 @@ export default class ReleaseBinaryCommand extends AppCommand {
...
+    // load release notes file or use provided release notes if none was specified
+    const releaseNotesString = await this.getReleaseNotesString();
 
-    debug("Creating release upload");
-    const createdReleaseUpload = await this.createReleaseUpload(client, app);
-    const uploadUri = createdReleaseUpload.uploadUrl;
-    const uploadId = createdReleaseUpload.uploadId;
+    // get number of distribution group users (and check distribution group existence)
+    // return null if request has failed because of any reason except non-existing group name.
+    const distributionGroupUsersNumber = await this.getDistributionGroupUsersNumber(client);
```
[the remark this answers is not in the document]
buptkang: Yeah, rollback to the original print result.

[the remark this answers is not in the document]
buptkang: Fixed.

buptkang:
Close this PR, will create another one to map a special release version.

PR #386, opened 2018-02-10
Make profile directory configurable

yuya-takeyama:
I'd like to `appcenter-cli` from Jenkins to automate CodePush deploy.
But it fails because of permission of `$HOME` (`/var/lib/jenkins`).

So I made the profile directory configurable with environment variable `APPCENTER_PROFILE_DIR`.

I think it's helpful to automate some tasks besides CodePush.
What do you think?

### DIFF
--- a/src/util/misc/get-profile-dir.ts
+++ b/src/util/misc/get-profile-dir.ts
@@ -7,6 +7,10 @@ import { profileDirName, oldProfileDirName, profileFile } from "./constants";
 const debug = require("debug")("appcenter-cli:util:misc:get-profile-dir");
 
 export function getProfileDir(): string {
+  if (typeof process.env.APPCENTER_PROFILE_DIR === 'string') {
+    return process.env.APPCENTER_PROFILE_DIR;
+  }
+
   const profileDir = path.join(getProfileDirParent(), profileDirName);
   const oldProfileDir = path.join(getProfileDirParent(), oldProfileDirName);
   if (!existsSync(profileDir) && existsSync(oldProfileDir)) {


christav:
Interesting idea, but I don't think that this is the right approach for automation - the profile is per user, so if you have multiple jobs running on the same agent, they'll share the profile, which means that if you have different logins between jobs they will interfere with each other in a very difficult to debug way.

For automation you should instead create an explicit access token and pass it using the `--token` switch on each command. This means you shouldn't need the profile directory at all and the issues are avoided.

We have a task on the backlog to add an environment variable for the token so you don't need to give it explicitly on every command but it hasn't gotten to the top of the priority list yet.

yuya-takeyama:
@christav 
Thanks for the comment!

>For automation you should instead create an explicit access token and pass it using the --token switch on each command. This means you shouldn't need the profile directory at all and the issues are avoided.

But I don't think that the way to passing a token is the problem.
Actually, I already tried it and it failed by the error which I described.
Even with the `--token` option, it creates `profile.json` with `"tokenSuppliedByUser": true`.

christav:
Are you trying to pass `--token` to the `appcenter login` command when that happens?

If you're passing the `--token` switch you don't need to run login at all - you can just skip that step. If on the other hand you're running another command and hitting the issue then that's a bug that needs to be fixed. Can you tell me which it is?

yuya-takeyama:
Ah, I misunderstood about it...

I tried like `appcenter codepush deployment --token ***` and `appcenter codepush release-react --token ***`.
Both of them worked as expected!

Thanks for guiding me!
Let me close.

christav:
Cool, glad it worked the way it was supposed to. 😃

PR #389, opened 2018-02-15
Fix the logic to get server URL

yuya-takeyama:
## Problem

`appcenter codepush release-react` fails with an error in a case:

```
Releasing update contents to CodePush:

Error: TypeError: Cannot read property 'environment' of null
```

## Steps to Reproduce

1. Make sure that `appcenter-cli` is logged out
1. Run `appcenter codepush release-react` with `--token` and without `--env`

## Root Cause

This is because `getUser()` does not always return a `Profile`.
It returns `null` when there is no `profile.json` in `~/.appcenter-cli` (see the line 145):
https://github.com/Microsoft/appcenter-cli/blob/d93908f5ed9e8523e020473f23b63c8c437beb2b/src/util/profile/profile.ts#L140-L161

In that case, `getUser().environement` throws the error:

https://github.com/Microsoft/appcenter-cli/blob/d93908f5ed9e8523e020473f23b63c8c437beb2b/src/commands/codepush/lib/release-command-skeleton.ts#L106

## What I Did

I added `getServerUrl()` method to do it correctly.

* If `--env` is specified, it returns the URL of the specified environment
* If a `Profile` exists (in `~/.appcenter-cli/profile.json`), it returns the URL of the the environment specified in `profile.environment`
  * If not, it returns the URL of `prod` env (as it's `defaultEnvironment`)

I also did some improvements for type definitions.

### DIFF
--- a/src/commands/codepush/lib/environment.ts
+++ b/src/commands/codepush/lib/environment.ts
@@ -39,6 +39,6 @@ const codePushEnvironmentsData: EnvironmentsFile = {
   }
 };
 
-export function environments(environmentName: string = codePushEnvironmentsData.defaultEnvironment): CodePushEnvironmentInfo {
+export function environments(environmentName: string = codePushEnvironmentsData.defaultEnvironment): CodePushEnvironmentInfo | undefined {
   return codePushEnvironmentsData.environments[environmentName];
 }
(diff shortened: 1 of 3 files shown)


Zakeelm:
@yuya-takeyama Hey going to get one of the CodePush Engineers to look at this! 😁

iTOYS:
@Zakeelm 
LGTM

PR #394, opened 2018-02-28
Add a new cli test parameter --vsts-id-variable <variable-name>

erychagov:


src/commands/test/lib/help-messages.ts
```diff
@@ -7,27 +7,27 @@ export module Messages {
       export const GenerateUITest = "Generates a Xamarin.UITest project";
 
       export const PrepareAppium = "Creates an artifacts directory with Appium tests.";
-      export const PrepareCalabash = `Creates an artifacts directory with Calabash tests.${os.EOL}` +
```
owenniblock: @erychagov - these bits look like they didn't need to change? What are you using to edit the files? I'm wondering if your tooling is touching the whitespace in some unexpected way?

owenniblock:
I just don't think they've changed - so it would be good to not include that change in the PR if possible. Added a not on the code line.

src/commands/test/lib/help-messages.ts
erychagov: @owenniblock 
I use VSCode. He removes all excess white spaces. That's my setting.

PR #400, opened 2018-03-05
New parameters for working with test results

erychagov:
```
--test-output-dir <output-directory>
```
Path to the directory for test results. This parameter won't work when you run using --async because we have to wait for test results. The results will be saved in the working directory if the relative path is specified.

```
--merge-nunit-xml <output.xml>
```
Merge the xml files in to the <output.xml> file. This parameter won't work when you run using --async because we have to wait for test results.


src/commands/test/lib/help-messages.ts
```diff
@@ -84,6 +83,9 @@ export module Messages {
 
       export const XCUITestIpaPath = "Path to the *.ipa file with the XCUITest tests";
       export const XCUITestBuildDir = "Path to the build output directory (usually <project>/Build/Products/Debug-iphoneos)";
+
+      export const TestOutputDir = "Path to the directory for test results. This parameter won't work when you run using --async because we have to wait for test results. The results will be saved in the working directory if the relative path is specified";
+      export const MergeNUnitXaml = "Merge the xml files in to the <output.xml> file. This parameter won't work when you run using --async because we have to wait for test results";
```
krukow: should this be: `MergeNUnitXaml` =>  `MergeNUnitXml` ?

src/commands/test/run/uitest.ts
```diff
@@ -64,10 +71,10 @@ export default class RunUITestsCommand extends RunTestsCommand {
   @hasArg
   excludeCategory: string[];
 
-  @help(Messages.TestCloud.Arguments.NUnitXml)
-  @longName("nunit-xml")
+  @help(Messages.TestCloud.Arguments.MergeNUnitXaml)
```
krukow: `MergeNUnitXml`

owenniblock:
Current review results:
 - [x] `npm build` successful
 - [ ] `npm run test` (fails):

```
> appcenter-cli@1.0.11 test /Users/owenniblock/appcenter-cli
> mocha --compilers ts:ts-node/register --recursive ./test/**/*-test.[tj]s

sh: mocha: command not found
npm ERR! file sh
npm ERR! code ELIFECYCLE
npm ERR! errno ENOENT
npm ERR! syscall spawn
npm ERR! appcenter-cli@1.0.11 test: `mocha --compilers ts:ts-node/register --recursive ./test/**/*-test.[tj]s`
npm ERR! spawn ENOENT
npm ERR! 
npm ERR! Failed at the appcenter-cli@1.0.11 test script.
npm ERR! This is probably not a problem with npm. There is likely additional logging output above.
```

src/commands/test/lib/help-messages.ts
```diff
@@ -7,27 +7,27 @@ export module Messages {
       export const GenerateUITest = "Generates a Xamarin.UITest project";
 
       export const PrepareAppium = "Creates an artifacts directory with Appium tests.";
-      export const PrepareCalabash = `Creates an artifacts directory with Calabash tests.${os.EOL}` + 
-                                     `Required external tools:${os.EOL}` +
```
owenniblock: Can we try to work out why your dev system keeps touching files like this when there are no changes? What are you using to edit files?

src/commands/test/run/uitest.ts
```diff
@@ -1,10 +1,17 @@
+import * as pfs from "../../../util/misc/promisfied-fs";
+import * as fs from "fs";
+import * as os from "os";
+import * as path from "path";
+import * as unzip from "unzip";
+import * as xmlUtil from "../../../util//misc/xml";
```
owenniblock: Remove double slash here.

src/commands/test/run/uitest.ts
```diff
@@ -119,4 +125,47 @@ export default class RunUITestsCommand extends RunTestsCommand {
...
     return this.buildDir;
   }
+
+  protected async mergeTestArtifacts(): Promise<void> {
+    if (!this.mergeNUnitXml) {
+      return;
+    }
+
+    let tempPath: string = await pfs.mkTempDir("appcenter-uitestreports")
+    let reportPath: string = path.join(this.generateReportPath(), "nunit_xml_zip.zip");
```
owenniblock: What happens here if the `generateReportPath()` returns `null` or a non existent path?

owenniblock: Also - can we check this on Windows to make sure it behaves the same?

src/util/apis/generated/models/index.d.ts
```diff
@@ -8431,6 +8432,18 @@ export interface TestReportStats {
...
   devicesSkipped: number;
   stepCount: number;
+  artifacts?: TestArtifacts;
+}
+
+/**
+ * @class
+ * Initializes a new instance of the TestArtifacts class.
+ * @constructor
+ * @member {string} nunit_xml_zip
```
owenniblock: I think this is autogenerated? So I guess there's nothing we can do to make this more generic? I.e. so that when other frameworks add their equivalents to the `nunit_xml_zip` key we don't have to regenerate this Interface?

owenniblock:
I'm also getting this warning when running the test:
> (node:70946) [DEP0013] DeprecationWarning: Calling an asynchronous function without callback is deprecated.
Can we fix this?

src/util/apis/generated/models/index.d.ts
erychagov: > I think this is autogenerated?

No, I've added it manually

owenniblock: I think this is meant to be autogenerated:

> Autorest-generated client code for talking to Bifrost.

Although I'm not sure what the process is for generating. Let me see if I can work it out - if not I'll see if @christav can point us in the right direction.

owenniblock: OK - you should be able to run this command to generate the model: `npm run autorest` just testing it now.

owenniblock: Looks like this removes the reference to the `artifacts` so I think we need to find a way to add this to the Bifrost API swagger so that the auto generation routine works correctly (otherwise another dev will remove this change next time the API code is regenerated).

erychagov:
@owenniblock 

> I'm also getting this warning when running the test:
>  ```(node:70946) [DEP0013] DeprecationWarning: Calling an asynchronous function without callback is deprecated.```
> Can we fix this?

Done

src/util/apis/generated/models/index.d.ts
owenniblock: @erychagov - I'm looking in to this now. I've put in a PR to update the Swagger YAML file. I've made it slightly more generic than your example here so we may need to tweak the code once it's available (it'll be a Dictionary rather than referencing `nunit_xml_zip` by name).

src/util/misc/xml.ts
```diff
@@ -0,0 +1,121 @@
...
+  var nodes: Node[] = collectAllElements(xml, "test-suite");
+  nodes.forEach((node: Node) => {
+    if (countChildren(node) <= 1) {
+      node.parentNode.removeChild(node);
+    }
+  });
+}
+
+export function combine(xml1: Document, xml2: Document): Document {
+  combineTestResultsAttribute(xml1, xml2, "total");
```
owenniblock: @erychagov - would it be possible to add some tests for this new util file? I'd be particularly interested in seeing how it behaves when the report xml files are malformed (e.g. missing expected elements etc).

Depends a little how long that would take so please push back if you don't think this would be valuable.

owenniblock: @erychagov - would it be possible to add some tests for this new util file? I'd be particularly interested in seeing how it behaves when the report xml files are malformed (e.g. missing expected elements etc).

Depends a little how long that would take so please push back if you don't think this would be valuable.

src/util/apis/generated/models/index.d.ts
christav: No manual changes to these files will be merged. These have to be autogenerated.

src/util/misc/promisfied-fs.ts
```diff
@@ -173,9 +173,24 @@ export async function walk(dir: string): Promise<string[]> {
   }
 }
 
+export function createLongPath(target: string) {
```
christav: Why is this completely synchronous function in a file called "promisified-fs"?

Going forward I want to get rid of this file, current versions of node support a native Promisify method that should easily replace all these wrappers. But that's for later.

src/util/misc/xml.ts
christav: This is a ton of test cloud specific stuff, not something that would be used across the entire CLI. It doesn't belong under util/misc, it should be in with your test cloud commands.

erychagov: @owenniblock 
I've added tests

src/util/apis/generated/models/index.d.ts
owenniblock: For some reason the property we need wasn't in the public API swagger yaml. Just waiting for that to be sorted so we can autogenerate.

src/commands/test/lib/run-tests-command.ts
```diff
@@ -248,8 +267,32 @@ export abstract class RunTestsCommand extends AppCommand {
...
+
+  private async downloadArtifacts(testRunId: string, artifacts: TestArtifacts): Promise<void> {
+    for (var key in artifacts) {
+
+      var reportPath = this.generateReportPath();
+      fsHelper.createLongPath(reportPath);
+      await downloadUtil.downloadFileAndSave(artifacts[key], path.join(reportPath, `${key.toString()}.zip`));
+
+      this.streamingOutput.text((command: RunTestsCommand): string => {
+        return `##vso[task.setvariable variable=${key}]${testRunId}${os.EOL}`;
```
owenniblock: Only just noticed this - the value should be the path to the .zip file (e.g. path.join(reportPath, `${key.toString()}.zip`))

rmarinho:
can I get a .exe of this to test ?

owenniblock:
@christav - I've updated the Swagger client code from local dev (with a fix in for that missing `artifacts` node) but it's also added a _ton_ of other stuff which suggests that the swagger generate hasn't been run for a while (and the API swagger may have changed without changing version which is different problem).

I'm a bit worried that this will have unforeseen consequences. Do we have a set of integration type tests we can run when this PR is ready (before we merge)?

src/util/misc/promisfied-fs.ts
owenniblock: @erychagov can we make this function `async`? Or has it been replaced by `fs-helper.ts`? Have we had a bad merge (may be my fault?) I'm sure you sorted this previously.

src/util/misc/xml.ts
owenniblock: @erychagov - I think this was deleted previously. but added back in by a bad merge. Or maybe it was never deleted - but the former seems more likely. We should delete this one?

erychagov:
@christav 
Could you please review the PR?

src/commands/test/lib/help-messages.ts
christav: Actually, it looks like we had some files previously checked in with trailing whitespace. This is a good thing - I really need to get TSLint hooked up to this thing.

src/commands/test/lib/junit-xml-util.ts
```diff
@@ -0,0 +1,174 @@
...
+import { DOMParser } from "xmldom";
+
+export class JUnitXmlUtil extends XmlUtil {
+
+  async mergeXmlResults(pathToArchive: string): Promise<Document> {
+    let tempPath: string = await pfs.mkTempDir("appcenter-junittestreports")
+    let mainXml: Document = this.getEmptyXmlDocument();
+
+    let self = this;
+    return new Promise<Document>((resolve,reject) => {
```
christav: Consider hooking up some 'error' event handlers on these pipes to properly detect if something goes wrong.

PR #402, opened 2018-03-06
Wrong android output directory detection fixed

mlegenhausen:
When you run `appcenter codepush release-cordova` before you have run `cordova platform add` the wrong output folder is detected under `android`. This pull request moves the detection after the `cordova build` command. In newer versions of the `cordova-cli` the missing platform is automatically added, so a manuell execution of `cordova platform add` is not needed.

I also improved the `getOutputFolder` function in that way that when no output folder at all could be detected an error is presented to the user.


iTOYS:
LGTM

buptkang:
👍

PR #403, opened 2018-03-08
Windows EPERM: Warn and don't bail on failed cleanup

krukow:
We've observed errors like:

```
Error: EPERM: operation not permitted, unlink 'C:\BuildAgent\temp\buildTmp\appcenter-upload118115-6800-1ohlguf.j77hk\AndroidTestServer.apk'
```

particularly on Windows. 

This should not cause the CLI to exit, but rather it should issue a warning.

### DIFF
--- a/src/commands/test/lib/run-tests-command.ts
+++ b/src/commands/test/lib/run-tests-command.ts
@@ -218,7 +218,7 @@ export abstract class RunTestsCommand extends AppCommand {
 
   protected async cleanupArtifactsDir(artifactsDir: string): Promise<void> {
     await pfs.rmDir(artifactsDir, true).catch(function(err){
-      console.warn(`Error ${err} while cleaning up artifacts directory ${artifactsDir}. Continuing without cleanup.`);
+      console.warn(`Error ${err} while cleaning up artifacts directory ${artifactsDir}. This is often due to files being locked or in use. Please check your virus scan settings and any local security policies you might have in place for this directory. Continuing without cleanup.`);
     });
   }
 
--- a/src/commands/test/lib/run-tests-command.ts
+++ b/src/commands/test/lib/run-tests-command.ts
@@ -217,7 +217,9 @@ export abstract class RunTestsCommand extends AppCommand {
   }
 
   protected async cleanupArtifactsDir(artifactsDir: string): Promise<void> {
-    await pfs.rmDir(artifactsDir, true);
+    await pfs.rmDir(artifactsDir, true).catch(function(err){
+      console.warn(`Error ${err} while cleaning up artifacts directory ${artifactsDir}. Continuing without cleanup.`);
+    });
   }
 
   private artifactsDir: string;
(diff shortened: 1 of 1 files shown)


krukow:
@owenniblock can you help me test this on windows?

owenniblock:
@krukow - sure. Do you have a surefire way of replicating? I would guess we can open one of the files that'll get cleaned up so it gets blocked.

krukow:
@owenniblock I don't - I was leaning on you to help :)

src/commands/test/lib/run-tests-command.ts
```diff
@@ -217,7 +217,9 @@ export abstract class RunTestsCommand extends AppCommand {
   }
 
   protected async cleanupArtifactsDir(artifactsDir: string): Promise<void> {
-    await pfs.rmDir(artifactsDir, true);
+    await pfs.rmDir(artifactsDir, true).catch(function(err){
+      console.warn(`Error ${err} while cleaning up artifacts directory ${artifactsDir}. Continuing without cleanup.`);
```
owenniblock: Would it make sense to try to lead the user here? If files aren't being cleaned up - we could get them to check virus scan and local permissions issues (or anything else that we've discovered causes this in Intercom)?

owenniblock:
No problem - let me fire up the Windows machine and see if I can replicate on the current CLI.

src/commands/test/lib/run-tests-command.ts
krukow: Absolutely - could you help me with some text?

owenniblock: How about:
```
console.warn(`Error ${err} while cleaning up artifacts directory ${artifactsDir}. This is often due to files being locked or in use. Please check your virus scan settings and any local security policies you might have in place for this directory. Continuing without cleanup.`);
```

krukow: 👍

owenniblock:
OK - I can replicate this by locking one of the .dll files as `read only` before the test completes.

I think we need to make a slight tweak to the message (we say `Error` way too much!) Maybe we should change it so that the description and help test comes before the error message which is a bit nonsensical?

Here's the error before this PR:
> Error: EPERM: operation not permitted, unlink 'C:\Users\ADMINI~1\AppData\Local\Temp\appcenter-upload11828-9124-cy0d4h.jrn48\Xamarin.UITest.dll'
> Error: There were Test Failures.

Here it is after:
> Error Error: EPERM: operation not permitted, unlink 'C:\Users\ADMINI~1\AppData\Local\Temp\appcenter-upload11828-4712-1498xu0.w4cy\Xamarin.UITest.dll' while cleaning up artifacts directory C:\Users\ADMINI~1\AppData\Local\Temp\appcenter-upload11828-4712-1498xu0.w4cy. This is often due to files being locked or in use. Please check your virus scan settings and any local security policies you might have in place for this directory. Continuing without cleanup.
> Error: There were Test Failures.

krukow:
@owenniblock `Error: There were Test Failures.`is this preventing the test run/scheduling? it shouldn't

krukow:
Number of times we say `Error` addressed

krukow:
Great - adding @christav for final review and ideally merge

krukow:
Merging: cc @christav FYI

PR #409, opened 2018-03-19
pass option all the way to the backend

john7doe:
Fix fixture chunking: The option is now included when starting the test run.

### DIFF
--- a/src/commands/test/run/uitest.ts
+++ b/src/commands/test/run/uitest.ts
@@ -71,12 +71,10 @@ export default class RunUITestsCommand extends RunTestsCommand {
 
   @help(Messages.TestCloud.Arguments.TestChunk)
   @longName("test-chunk")
-  @hasArg
   testChunk: boolean;
 
   @help(Messages.TestCloud.Arguments.FixtureChunk)
   @longName("fixture-chunk")
-  @hasArg
   fixtureChunk: boolean;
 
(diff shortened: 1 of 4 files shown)


amchew:
Hi @john7doe , thanks for submitting the PR ! We've set up a [process](https://msmobilecenter.visualstudio.com/Mobile-Center/_wiki/wikis/Mobile-Center.wiki?wikiVersion=GBwikiMaster&pagePath=%2FCustomer%20Service%20and%20Support%2FGitHub%20Support%20Process%2FSDK%20and%20CLI%20Repositories%2FContributions%20to%20a%20repository) on how to contribute to the CLI repo, and now, PR submitters should include a description of the change they would like to include in the changelog. Each time a PR is merged and the next version of the CLI is released, the first paragraph in the PR description will be copied into the changelog. Could you include a description of the change in the PR description?

src/commands/test/run/uitest.ts
```diff
@@ -116,6 +118,15 @@ export default class RunUITestsCommand extends RunTestsCommand {
     return preparer.prepare();
   }
 
+  protected getParametersFromOptions() : {} {
```
owenniblock: I think that on the prepare - the necessary `chunker` stuff will be handled by shelling out to the `test-cloud.exe` - but wanted to double check that this was the case.

PR #410, opened 2018-03-19
Add error message if server message is empty

erychagov:
Print an useful error message instead of an empty one.

For some reason server side returns the empty error message. It has been replaced by a general message which says "Could not start your tests. Maybe your subscription has expired."

### DIFF
--- a/src/commands/test/lib/run-tests-command.ts
+++ b/src/commands/test/lib/run-tests-command.ts
@@ -182,6 +182,9 @@ export abstract class RunTestsCommand extends AppCommand {
       }
       else
       {
+        if(!err.message) {
+          err.message = "Could not start your tests. Maybe your subscription has expired.";
+        }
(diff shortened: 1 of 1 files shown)


amchew:
Hi @erychagov , thanks for submitting the PR ! We've set up a [process](https://msmobilecenter.visualstudio.com/Mobile-Center/_wiki/wikis/Mobile-Center.wiki?wikiVersion=GBwikiMaster&pagePath=%2FCustomer%20Service%20and%20Support%2FGitHub%20Support%20Process%2FSDK%20and%20CLI%20Repositories%2FContributions%20to%20a%20repository) on how to contribute to the CLI repo, and now, PR submitters should include a description of the change they would like to include in the changelog. Each time a PR is merged and the next version of the CLI is released, the first paragraph in the PR description will be copied into the changelog. Could you include a description of the change in the PR description?

PR #415, opened 2018-03-28
Download tests report

erychagov:
The new appcenter-cli command `download` has been created to download the report artifacts, unpack and merge them.

The appcenter-cli can be called in `--async` mode whereby it sends the tests and drops the connection before the tests complete.
This is useful in certain CI system (e.g. VSTS) where holding up the build step can cause the polling of the appcenter-cli to time out.


owenniblock:
@christav - this was the PR I was referring to. I'd like to get this out next week if we can. I can start work on mapping out the VSTS task stuff once this is in.

Please let me know if you see anything wonky and we'll sort it out ASAP.

krukow:
@erychagov @owenniblock I've added a  few comments

owenniblock:
@krukow I'll take a look at this later. I think it would be better to use a more generic form of `--merge-nunit-xml` like we are in the download command?

Something like `--merge-report-file` which merges the report file where available?

so in NUnit we will pick up the nunit artifact zip and process appropriately but we can also use the same parameter for JUnit it could (with a bit of intelligent monkeying) _just work_

krukow:
@owenniblock you're right it should be a more generic form

owenniblock:
@erychagov - I didn't get chance to do this work this evening. I think we need to add the following parameter to `appcenter test run manifest`

`--merge-report-file <file-name>` when set, it should check for the artefact (either junit_xml_zip or nunit_xml_zip) and perform the relevant merge operation.

~~~~~~~~~~~

@krukow - if I'm understanding you correctly on this:
> merged-file-name is an optional parameter. If called without merged-file-name, the tool should merely download and stores files in the test-output-dir

We should not attempt to merge the files - just download and unzip in to the folder?

krukow:
@owenniblock you shouldn't even assume that artifacts are .zip files. Merely download the artifacts to the `test-output-folder`

krukow:
Other than that - yes :)

src/commands/test/run/manifest.ts
```diff
@@ -33,4 +42,33 @@ export default class RunManifestTestsCommand extends RunTestsCommand {
...
+    files.every((file: string) => {
+      if (path.extname(file) === ".zip") {
+        xmlUtil = XmlUtilBuilder.buildXmlUtilByString(path.basename(file));
+        return false;
+      }
+      return true;
+    });
+
+    if (!xmlUtil) {
+      throw new Error("Arficats folder doesn't contain an archive with test results");
```
krukow: @erychagov there's a spelling error here:
I propose: `"test-output-dir" doesn't contain any mergeable test results (.zip archives containing .xml documents)`

src/util/misc/download.ts
```diff
@@ -22,3 +26,20 @@ export function downloadFileAndSave(downloadUrl: string, filePath: string): Prom
...
+export async function downloadArtifacts(command: AppCommand, streamingOutput: StreamingArrayOutput, outputDir: string, testRunId: string, artifacts: { [propertyName: string]: string }): Promise<void> {
+  for (const key in artifacts) {
+
+    const reportPath: string = fsHelper.generateAbsolutePath(outputDir);
+    const pathToArchive: string = path.join(reportPath, `${key.toString()}.zip`);
+    fsHelper.createLongPath(reportPath);
+    await downloadFileAndSave(artifacts[key], pathToArchive);
+
+    // Print only in VSTS environment
+    if (process.env["TF_BUILD"]) {
```
krukow: is `TF_BUILD` always present in VSTS and TFS?  IS there a document that describes this somewhere?

krukow:
@erychagov I've added a few comments inline .

Can you let me know which part of the `--merge-nunit-xml` solution you're unhappy about?

@rmarinho do you have any input?

src/commands/test/run/manifest.ts
```diff
@@ -33,4 +42,33 @@ export default class RunManifestTestsCommand extends RunTestsCommand {
...
+
+  protected async mergeTestArtifacts(): Promise<void> {
+    if (!this.outputXmlName) {
+      return;
+    }
+
+    const files: string[] = await walk(this.testOutputDir);
+    let xmlUtil: XmlUtil = null;
+    files.every((file: string) => {
+      if (path.extname(file) === ".zip") {
```
erychagov: @krukow 
I don't like this place.
Is it possible to have few archives inside test artifacts? In this case the behaviour is incorrect. If the correct archive is not first we get an exception `Unexpected reports type`. Also we get this exception in case when if artifacts don't contain any archive.

So with current realization we'll never get the exception:
> "test-output-dir" doesn't contain any mergeable test results (.zip archives containing .xml documents)

src/util/misc/download.ts
erychagov: @krukow 
https://docs.microsoft.com/en-us/vsts/build-release/concepts/definitions/build/variables?view=vsts&tabs=batch#tfbuild

Also I've asked about it in a channel related to VSTS

krukow: @erychagov great - please add the link to the code comment

src/commands/test/run/manifest.ts
krukow: @erychagov I'd probably write it so that it expects exactly one `.zip` file when `this.outputXmlName` is set. It ignores all other files, expects exactly one .zip file which is then merged. Raises errors if there is fewer or more than one?

erychagov: @krukow 
The code raises the error `Unexpected reports type` if the first (or single) archive has an incorrect name.
If there is not any archive the code raises the error `"test-output-dir" doesn't contain any mergeable test results (.zip archives containing .xml documents)`. I was wrong when I said `we'll never get the exception`

> I'd probably write it so that it expects exactly one .zip file

Ok, I'll do it

krukow:
@owenniblock and @erychagov I'd like you to test this thoroughly, and ideally merge tomorrow. 

Then please do a PR which adds the CLI version number to `--debug` output

Then please ask @christav to support us in making a release

krukow:
Also add the CLI version to the app upload (parameter in the manifest).

owenniblock:
My testing:
- [x] Builds locally
- [x] `npm test` runs
- [x] Local run with debug outputs cli version
- [ ] Uploads manifest containing cli version
- [ ] Uploads and runs tests for all frameworks correctly
- [ ] Uploads, runs and returns NUnit XML for UITest framework
- [ ] Uploads, runs and returns JUnit XML for Appium framework

owenniblock:
@krukow NOTE: I don't think the JUnit output is currently on the active pipeline branch. Can we confirm? Without this the JUnit bits won't work.

krukow:
Very nice work @owenniblock @erychagov

PR #433, opened 2018-05-01
Allow debug uitest console to be used when calling appcenter-cli

owenniblock:
### Motivation
To allow the UITest CI to use the dev version of the test-cloud.exe file - we need to allow it in the CLI.

This small code change allows this (along with some changes in UITest itself).

### DIFF
--- a/src/commands/test/lib/uitest-preparer.ts
+++ b/src/commands/test/lib/uitest-preparer.ts
@@ -194,7 +194,7 @@ export class UITestPreparer {
     const testCloudPathDev = path.join(toolsDir, "Xamarin.UITest.Console.exe");
 
     if (!await fileExists(testCloudPath) && !await fileExists(testCloudPathDev)) {
-      throw new Error(`Cannot find test-cloud.exe.., the exe was not found in the path specified by "--uitest-tools-dir".${os.EOL}` +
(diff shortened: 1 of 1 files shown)


src/commands/test/lib/uitest-preparer.ts
```diff
@@ -191,13 +191,18 @@ export class UITestPreparer {
...
+    const testCloudPathDev = path.join(toolsDir, "Xamarin.UITest.Console.exe");
 
-    if (!await fileExists(testCloudPath)) {
+    if (!await fileExists(testCloudPath) && !await fileExists(testCloudPathDev)) {
       throw new Error(`Cannot find test-cloud.exe, the exe was not found in the path specified by "--uitest-tools-dir".${os.EOL}` +
         `Please check that ${testCloudPath} points to a test-cloud.exe.${os.EOL}` +
         `Minimum required version is "${this.getMinimumVersionString()}".`);
     }
 
+    if (!await fileExists(testCloudPath)) {
```
erychagov: @owenniblock 
Double checking `!await fileExists(testCloudPath)`
I would avoid it.

owenniblock: Happy for you to change it if there's a better way - I was just copying the existing pattern.

PR #434, opened 2018-05-03
Mark the option `--dsym-dir` as deprecated

erychagov:
We don't support showing symbolicated crashes.

The option `--dsym-dir` is deprecated and ignored.

The functionality with the ability to view and download raw (unsymbolicated) crash reports is available on appcenter.ms

### DIFF
--- a/src/commands/test/lib/help-messages.ts
+++ b/src/commands/test/lib/help-messages.ts
@@ -45,7 +45,7 @@ export module Messages {
 
       export const PrepareArtifactsDir = "Path to the artifacts directory to create";
       export const RunDevices = "Device selection slug";
-      export const RunDSymDir = "Path to the directory with iOS symbol files";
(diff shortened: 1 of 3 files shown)


kamstrup:
We can not break command line API like this. appcenter-cli is public API for all intents and purposes and is as such set in stone IMHO. We should still allow the parameter, but log a warning that it is deprecated and ignored.

owenniblock:
@jmoody @krukow - If we're def making this available again - maybe we could create a VSTS task for the work and link to this PR so the engineer assigned just has to undo the commit when it's ready to be reintroduced?

Looks good to me though.

jmoody:
dSYM parsing will probably not be available.

nathanielcook:
Why are crashes that occur during test runs not symbolicated? This is something that was available with Xamarin Test Cloud.

Issue #442, opened 2018-05-18
URLs generated by "test run" command are not escaped


jmoody:
### Reproduce

Requires:

* macOS
* Xcode
* local clone of https://github.com/Microsoft/AppCenter-Test-XCUITest-Extensions/
* add Flowers iOS to your appcenter account
* generate a device set

```
$ bin/make/build-for-testing.sh arm
$ SERIES="Spaces In Name" bin/make/appcenter.sh < device set >
```

### Expected

`appcenter appcenter test run` to print a valid URL to stdout at the end of the job.

### Found

The URL is invalid because it contains spaces.

Example:

```
https://appcenter.ms/orgs/org/apps/Flowers/test/series/DYLD Bootstrap.dylib/runs/udid
```

### Notes

This is a problem for CI systems that attempt to inject this URL.

amchew:
Hi @jmoody, an update, we're merged the fix in the [PR](https://github.com/Microsoft/appcenter-cli/pull/443) and will be releasing the fix soon.

amchew:
Hi @jmoody this issue should have been fixed with the PR #443, and released in the new CLI v1.1.0 version today. Could you install the new CLI version, and check if this issue is resolved? Thanks.

amchew:
Hi @jmoody this issue should have been fixed with the PR #443, please test it out. Feel free to reach out again if you have any questions! I'll be closing this ticket for now.

Amanda

PR #443, opened 2018-05-18
Add url encoding

erychagov:
This PR fixes the issue #442

### DIFF
--- a/test/commands/test/util/portal-helper-test.ts
+++ b/test/commands/test/util/portal-helper-test.ts
@@ -1,0 +1,21 @@
+import { getPortalTestLink } from "../../../../src/util/portal/portal-helper";
+import { expect } from "chai";
+
+describe("portal helper test", function () {
+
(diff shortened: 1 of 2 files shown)


test/commands/test/util/portal-helper-test.ts
```diff
@@ -0,0 +1,21 @@
...
+  it("should escape url for orgs", () => {
+    const url: string = generateUrl(true);
+    const spaceIndex: number = url.indexOf(" ");
+    expect(spaceIndex).to.eql(-1);
+  });
+
+  it("should escape url for users", () => {
+    const url: string = generateUrl(false);
+    const spaceIndex: number = url.indexOf(" ");
+    expect(spaceIndex).to.eql(-1);
```
owenniblock: I think we should be checking that %20 exists the expected number of times here. Otherwise good - minor so going to approve anyway :)

PR #444, opened 2018-05-18
Add a new version check

erychagov:
Having an update check running in the command printing a friendly reminder could maybe help few users and prevent few support cases.

### DIFF
--- a/src/util/commandline/command.ts
+++ b/src/util/commandline/command.ts
@@ -115,11 +115,16 @@ export class Command {
       setQuiet();
     }
 
-    // Check if a new cli version is available
-    const result: Buffer = execSync("npm show appcenter-cli version");
-    const newVersion: string = result.toString("utf-8", 0, result.length - os.EOL.length);
-    if (newVersion !== this.getVersion()) {
-      console.log(`The new version is available: ${newVersion}`);
+    try {
+      // Check if a new cli version is available
+      const result: Buffer = execSync("npm show appcenter-cli version");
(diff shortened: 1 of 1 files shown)


src/util/commandline/command.ts
```diff
@@ -113,6 +115,13 @@ export class Command {
       setQuiet();
     }
 
+    // Check if a new cli version is available
+    const result: Buffer = execSync("npm show appcenter-cli version");
+    const newVersion: string = result.toString("utf-8", 0, result.length - os.EOL.length);
+    if (newVersion !== this.getVersion()) {
+      console.log(`The new version is available: ${newVersion}`);
```
owenniblock: Change the wording to:
`A new version of appcenter-cli is available: ${newVersion}`

src/util/commandline/command.ts
```diff
@@ -113,6 +115,13 @@ export class Command {
       setQuiet();
     }
 
+    // Check if a new cli version is available
+    const result: Buffer = execSync("npm show appcenter-cli version");
```
owenniblock: We should catch and output any errors here - if something goes wrong checking the version - we don't want to stop the cli from working.

christav:
Rather than doing this "manually" I'd rather do this via this package: https://www.npmjs.com/package/update-notifier

This is the same package that npm itself uses so we know it'll be kept up to date based on what NPM does. That and I'd rather have this check in the CLI entry point rather than in the commandline class.

christav:
Closing in favor of #458

PR #449, opened 2018-06-11
Add validation to uploaded .dll.config files to avoid UITest stalls

owenniblock:
During test run execution - when a file ending in .dll.config is uploaded and it's not a valid XML file, we add a warning to the console output and don't send that file.

When UITest files are uploaded - if they include empty .dll.config files (or malformed files) it stalls the test.

This work filters out these files and shows the user a warning message.


src/commands/test/lib/included-files-parser.ts
```diff
@@ -1,8 +1,32 @@
...
+    return;
+  }
+
+  const includedFiles = parseIncludedFiles(include, rootDir);
+
+  for (let i = 0; i < includedFiles.length; i++) {
+    const includedFile = includedFiles[i];
+    const copyTarget = path.join(path.dirname(rootDir), includedFile.targetPath);
+
+    if (copyTarget.indexOf(".dll.config") !== -1 && !validateXmlFile(copyTarget)) {
```
christav: this test will match "x.dll.config" and ".dll.config.x" and "a.b.dll.config.c.txt". I'd suggest using the lodash function `_.endsWith(".dll.config")` instead - only matches at the end and is more explicit about what you're checking for.

Unless you wanted all the other matches, in which case carry on. 😄

src/commands/test/lib/prepare-tests-command.ts
```diff
@@ -74,18 +73,7 @@ export class PrepareTestsCommand extends Command {
...
-
-    const includedFiles = parseIncludedFiles(this.include, this.getSourceRootDir());
-    for (let i = 0; i < includedFiles.length; i++) {
-      const includedFile = includedFiles[i];
-      const copyTarget = path.join(this.artifactsDir, includedFile.targetPath);
-      await pfs.cp(includedFile.sourcePath, copyTarget);
-
-      manifest.files.push(includedFile.targetPath);
-    }
+    copyIncludedFiles(manifest, this.include, this.getSourceRootDir());
```
christav: Nice removal of duplication!

src/commands/test/lib/included-files-parser.ts
```diff
@@ -1,8 +1,32 @@
-import { IFileDescriptionJson } from "./test-manifest-reader";
+import { IFileDescriptionJson, ITestCloudManifestJson } from "./test-manifest-reader";
 import * as path from "path";
+import * as pfs from "../../../util/misc/promisfied-fs";
+import { validateXmlFile } from "./xml-util";
+import { out } from "../../../util/interaction";
 
 const invalidCharactersRegexp = /['"!#$%&+^<=>`|]/;
 
+export async function copyIncludedFiles(manifest: ITestCloudManifestJson, include: string[], rootDir: string) {
```
christav: Are there any tests for this stuff?

owenniblock: They were sadly lacking on the original logic - I can add some if you need me to?

test/commands/test/lib/included-files-parser-test.ts
```diff
@@ -1,7 +1,112 @@
...
+    files: ["file1", "file2"], testFramework: { name: "", data: "" } };
+    const expectedManifest = { schemaVersion: "1.0.0", cliVersion: "1.0.0",  applicationFile: "",
+      files: ["file1", "file2"], testFramework: { name: "", data: "" } };
+
+    await copyIncludedFiles(inputManifest, [], null);
+
+    expect(inputManifest).to.deep.equal(expectedManifest);
+  });
+
+  it("should do nothing if something.dll.config exists in include", async () => {
```
john7doe: The is part does not mention that fact that something.dll.config does not contain valid xml.

Not sure what the right behaviour:

1) always exclude *.dll.config files that does not contain valid xml
2) only exclude *.dll.config files if they do not contain valid xml AND a .dll file exists.

Both strategies will avoid the error, but 1) is more restrictive excludes some "valid"/working case 

Seems like you are opting for 1)

test/commands/test/lib/included-files-parser-test.ts
```diff
@@ -1,7 +1,112 @@
...
+    const inputManifest = { schemaVersion: "1.0.0", cliVersion: "1.0.0",  applicationFile: "",
+    files: ["file1", "file2"], testFramework: { name: "", data: "" } };
+    const expectedManifest = { schemaVersion: "1.0.0", cliVersion: "1.0.0",  applicationFile: "",
+      files: ["file1", "file2"], testFramework: { name: "", data: "" } };
+
+    await copyIncludedFiles(inputManifest, null, null);
+
+    expect(inputManifest).to.deep.equal(expectedManifest);
+  });
+
```
john7doe: Consider using context, to capture that *dll.config can be valid/invalid (https://mochajs.org/#bdd):

test/commands/test/lib/included-files-parser-test.ts
```diff
@@ -1,7 +1,112 @@
...
+    sandbox = Sinon.sandbox.create();
+  });
+  afterEach(() => {
+      sandbox.restore();
+  });
+
+  it("should do nothing if null include is sent", async () => {
+    const inputManifest = { schemaVersion: "1.0.0", cliVersion: "1.0.0",  applicationFile: "",
+    files: ["file1", "file2"], testFramework: { name: "", data: "" } };
+    const expectedManifest = { schemaVersion: "1.0.0", cliVersion: "1.0.0",  applicationFile: "",
```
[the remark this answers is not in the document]
owenniblock: I have returned a manifest to make testing easier - however I have left it also updating the referenced manifest (I don't want to risk introducing regressions this late in the PR process).

I had to monkey about with the assignment of `inputManifest` to enable it to get the references object (setting the value in the before function).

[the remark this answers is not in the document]
john7doe: I think it should be one or the other:

1) return a modified manifest and do not modify the existing one.
2) leave it as it is.

I would actually prefer 2) over the current

"I don't want to risk introducing regressions this late in the PR process)."
The tests are there for us to detect a regression, hence be comfortable with make a change like the one I suggested.

With option 1) there would be no need for monkeying :-)

src/commands/test/lib/included-files-parser.ts
```diff
@@ -1,8 +1,36 @@
...
+    return manifest;
+  }
+
+  const includedFiles = this.parseIncludedFiles(include, rootDir);
+
+  for (let i = 0; i < includedFiles.length; i++) {
+    const includedFile = includedFiles[i];
+    const copyTarget = path.join(path.dirname(rootDir), includedFile.targetPath);
+
+    if (_.endsWith(copyTarget, ".dll.config") && (manifest.files.indexOf(includedFile.targetPath.slice(0, -7)) > -1
```
john7doe: either work on copyTarget or includedFile.targetPath, not both. Preferable includedFile.targetPath since the logic is deciding if it should be copied at all)  

The if statement is really hard to read and I'm not sure it is working correctly. copyTarget file will not exist, its first created/copied a few files down. (await pfs.cp(includedFile.sourcePath, copyTarget))

I think something like this is more readable: 
```
for (let i = 0; i < includedFiles.length; i++) {
  const includedFile = includedFiles[i];
 if (_.endsWith(includedFile.targetPath, ".dll.config") {
   const assemblyName = includedFile.targetPath.slice(0, -7);
   const hasCorrespondingAssembly = manifest.files.indexOf(assemblyName) > -1 || include.indexOf(assemblyName);
  if(hasCorrespondingAssembly && !validXmlFile(includedFile.targetPath) {
      out.text(`Warning: The XML config file ${copyTarget} was not a valid XML file. This file will not be uploaded.`);
      continue;
   }
  }
   const copyTarget = path.join(path.dirname(rootDir), includedFile.targetPath);   
   await pfs.cp(includedFile.sourcePath, copyTarget);
   manifest.files.push(includedFile.targetPath);
 }
```
Sorry for the formatting and compile error, it was typed in the browser, but hopefully understandable :-)

owenniblock: Good call! I don't know why I was using `copyTarget` in the first place - changing.

test/commands/test/lib/included-files-parser-test.ts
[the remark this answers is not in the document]
owenniblock: OK, I've moved it back to 2) purely because there are other places where we send the manifest as a reference - I agree that it would be better to clone and return the manifest but I don't want to have inconsistencies in how different areas behave - if I had more time I'd refactor all these areas so they work by returning a new copy of the manifest.

However, the refactoring does mean that I no longer need to reference the manifest at all to test my changes (the filter step is a distinct step as discussed).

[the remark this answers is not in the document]
owenniblock: > The tests are there for us to detect a regression, hence be comfortable with make a change like the one I suggested.

Worth noting that not all of this is my code and the project as a whole is light on tests so I'm naturally scared of making too many structural changes. I would _love_ to spend some time improving the test coverage and testability of the CLI though!

package.json
```diff
@@ -61,7 +61,7 @@
     "@types/request": "^2.0.3",
     "@types/rimraf": "^2.0.2",
     "@types/semver": "^5.3.33",
-    "@types/sinon": "^2.3.5",
```
john7doe: Would it not be better if the type definitions and the package itself has the same version? Last version I could find  is 5.0.1

src/commands/test/lib/included-files-parser.ts
```diff
@@ -1,12 +1,63 @@
...
+    manifest.files.push(includedFile.targetPath);
+  }
+}
+
 export function parseIncludedFiles(includedFiles: string[], rootDir: string): IFileDescriptionJson[] {
   return includedFiles.map((f) => parseIncludedFile(f, rootDir));
 }
 
+export function filterIncludedFiles(manifestFiles: string[], include: string[]): string[] {
+  if (!include) {
```
john7doe: Problematic files can also be in the manifest, so guess that test should really be on allIncludedFiles. But if allIncludedFiles are empty then the result would be [], so there is no need for the special handling at all.

src/commands/test/lib/included-files-parser.ts
```diff
@@ -1,12 +1,63 @@
...
+    return [];
+  }
+
+  const allIncludedFiles = manifestFiles.concat(include);
+  const output = [];
+  for (const file of include) {
+    if (validFile(file, allIncludedFiles)) {
+      output.push(file);
+    }
+  }
```
john7doe: lines 38-45 Could be replaced with `return allIncludedFiles.filter((f) => validFile(file, allIncludedFiles));`

src/commands/test/lib/included-files-parser.ts
owenniblock: Yes - you're right. I know that the `.dll.config` files have to be manually added using the `--include` flag.

I'm going to leave this for now but perhaps future development could include moving this filter to `allIncludedFiles`. Perhaps we could do this if we ever have need to filter out files that are naturally included in the manifest files? It would require some minor restructuring of the way it's working.

src/commands/test/lib/included-files-parser.ts
owenniblock: Done - need to be:
> return include.filter((f) => validFile(f, allIncludedFiles));

We're only filtering on `--include` on this PR - I made the assumption that we only choose files automatically if those files are OK to upload (that certainly seems to be the case when testing the CLI with commands).

package.json
owenniblock: _reset_ It broke the build for some reason.

PR #452, opened 2018-06-13
Custom telemetry source using '--telemetry-source' argument

evgeny-pol:
Allow applications to pass it's own telemetry source, when using appcenter-cli to access Visual Studio App Center

### DIFF
--- a/src/util/apis/create-client.ts
+++ b/src/util/apis/create-client.ts
@@ -23,10 +23,10 @@ export interface AppCenterClientFactory {
   fromProfile(user: Profile): AppCenterClient;
 }
 
-export function createAppCenterClient(command: string[], telemetryEnabled: boolean): AppCenterClientFactory {
+export function createAppCenterClient(command: string[], telemetryEnabled: boolean, telemetrySource: string): AppCenterClientFactory {
   function createClientOptions(): any {
     debug(`Creating client options, isDebug = ${isDebug()}`);
-    const filters = [userAgentFilter, telemetryFilter(command.join(" "), telemetryEnabled)];
+    const filters = [userAgentFilter, telemetryFilter(command.join(" "), telemetryEnabled, telemetrySource)];
     return {
       filters: isDebug() ? [createLogger()].concat(filters) : filters
     };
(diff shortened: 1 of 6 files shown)


christav:
Why are we doing this? What's the use case?

Doing this makes things harder on us on the back end - queries in our logs get harder since there's now an unbounded set of telemetry sources instead of just a couple. And that help text for the switch is completely inadequate - it'll be displayed in every command help screen and doesn't explain a thing about what this is actually FOR.

I really don't want to do this unless there's a really compelling reason I'm not seeing right now.

evgeny-pol:
Hello @christav , sorry, should've written a detailed description, to clarify, why we need this change.
This argument will allow other applications to use appcenter-cli as a mediator between itself and App Center portal.

More concretely, right now we are working on a App Center extension for Visual Studio Code:
https://github.com/Microsoft/vscode-appcenter

We mostly use autogenerated REST API client to issue commands to App Center, so we can set correct telemetry source ourself in those places:
https://github.com/Microsoft/vscode-appcenter/pull/108/files

But in one place, to launch UI tests, we use appcenter-cli:
https://github.com/Microsoft/vscode-appcenter/blob/master/src/tests/appCenterUITestRunner.ts#L112

Right now in a telemetry dashboard it will look like user launched UI tests from a appcenter-cli, but in reality he did it from a VS Code extension
If we start tests using appcenter-cli while passing it our telemetry source we will be able to easily see how our extension are used, by filtering telemetry by our telemetry source

I hope that I was able to show the reasoning behind this changes and I will be happy to hear your thoughts about how we can improve this solution

christav:
Since this is completely internal, rather than hacking up the command line parameter parsing I'd rather see this in an environment variable instead. Plus that way you don't need to set it on every command, but instead just set it in the environment for the shell you're running the CLI in. Much less intrusive, doesn't show up in the help, and gives you what you want.

evgeny-pol:
@christav Thanks for a great idea, we are going use it and implement passing of telemetry source using environment variable, so I am going to close this PR and open another one, when we get everything ready

PR #454, opened 2018-06-15
Get custom telemetry source from environment variable

evgeny-pol:
Allow applications to pass it's own telemetry source via environment variable TELEMETRY_SOURCE, when using appcenter-cli to access Visual Studio App Center

It will be used by "App Center extension for Visual Studio Code" when calling appcenter-cli to start user's application UI tests in App Center:
https://github.com/Microsoft/vscode-appcenter

This option is intended for internal usage so it should not be mentioned in release notes


src/util/apis/telemetry-filter.ts
```diff
@@ -15,11 +15,12 @@ const sessionHeaderName = "diagnostic-context";
 const commandNameHeaderName = "cli-command-name";
 
 export function telemetryFilter(commandName: string, telemetryIsEnabled: boolean) : {(resource: WebResource, next: any, callback: any): any} {
+  const telemetrySource = process.env.TELEMETRY_SOURCE || "cli";
```
christav: Please add a function to the src/util/profile/environment-vars.ts file to read this environment variable. That way all the evenironment variables we look for are in a single place in the code.

Also prefix the variable name with APPCENTER: `APPCENTER_TELEMETRY_SOURCE`.

evgeny-pol:
Hello @christav , I've made a new PR with the changes you requested, please check it out

christav:
@evgeny-pol Thanks, exactly what I was looking for.

PR #457, opened 2018-06-28
Bug 41992: CLI: NUnit report download intermittent failures (Xamarin.Forms)

AndreyMitsyk:
**Problems:**
- In result report we have "XML merging has ended with an error" without any additional information:
If something is wrong with XML, we will have following: for `nunit-xml-util` we will have unhanded exception in different thread when we try to use `*Transformation`. For `junit-xml-util` we will have `EmptyXmlDocument` without errors on top level.

- XML validation does not work correctly:
E.g. we have `validXmlFile` function that should return negative result for broken XML file. Unfortunately, result is always true. That happens because DOMParser does not trigger real JS exception and just send errors to console:
```
  [xmldom warning]	attribute "<test-suite" missed value!! "<test-suite" instead2!!
@#[line:2,col:1]
[xmldom error]	element parse error: Error: invalid attribute:<test-suite
@#[line:2,col:1]
```
In this case we have that xml document is not null, but this is incorrect document with parsererror nodes from DOMParser.

**Solutions:**
- Add DOMParser configuration with `errorHandler` to process XML errors and throw an exception.
- Added explicit rejection to see an error on top level.

**Changes:**
- Added mimeTypes for parseFromString inside *xml-util classes.
- Added explicit rejection for nunit-xml-util and junit-xml-util.
- Implemented logic to handle DOMParser errors.
- Fixed XML verification.
- Implemented tests for DOMParser error handling.
- Extended existing tests.
- Added test data.
- Merged changes regarding typescript compilation.


krukow:
@owenniblock please review

src/commands/test/lib/nunit-xml-util.ts
```diff
@@ -11,29 +11,45 @@ export class NUnitXmlUtil extends XmlUtil {
...
 
-    const self = this;
+    const self: NUnitXmlUtil = this;
     return new Promise<Document>((resolve, reject) => {
       fs.createReadStream(pathToArchive)
         .pipe(unzip.Parse())
         .on("entry", function (entry: unzip.Entry) {
-          const fullPath = path.join(tempPath, entry.path);
+          // Skip directories and hidden system files
+          if (entry.type === "Directory" || path.basename(entry.path).substring(0, 1) === ".") {
```
owenniblock: Is there anything we can do to refactor the shared code here and in `junit-xml-util`?

src/commands/test/lib/xml-util.ts
```diff
@@ -5,49 +5,56 @@ export abstract class XmlUtil {
...
+      result += this.countChildren(element.childNodes[i] as Element);
     }
     return result;
   }
 }
 
 export function validXmlFile(file: string): boolean {
   try {
-    const xml = new DOMParser().parseFromString(fs.readFileSync(file, "utf-8"));
+    const configuration: object = {
```
owenniblock: Can we share this configuration somewhere so it's defined in one place?

test/commands/test/resources/broken/nunit_report.xml
```diff
@@ -0,0 +1,117 @@
+<?xml version="1.0" encoding="utf-8"?>
```
owenniblock: It would be good for me to understand what causes this file to be broken - have we manually broken it? Or is this the example .xml as it was provided from the failing CLI run and we previously just weren't flagging up the issue?

test/commands/test/util/nunit-xml-util-test.ts
```diff
@@ -141,9 +141,25 @@ describe("nunit xml util", function () {
...
 
     // Then
-    const finalStrXml = new XMLSerializer().serializeToString(xml);
+    const finalStrXml: string = new XMLSerializer().serializeToString(xml);
 
     // Doesn't throw exception
     xmlLib.parseXml(finalStrXml);
   });
+
+  it("should throw an explicit exception", async () => {
```
owenniblock: Loving the additional tests! ❤️

test/commands/test/resources/broken/nunit_report.xml
AndreyMitsyk: It's a manually broken file

owenniblock: Part of me was hoping it was something from the intermittent issue we've been seeing 🙂

src/commands/test/lib/nunit-xml-util.ts
AndreyMitsyk: Will do this

src/commands/test/lib/xml-util.ts
AndreyMitsyk: I will refactor this as well

owenniblock:
- [This PR is work for this error](https://github.com/Microsoft/appcenter-cli/issues/447) - might not resolve the issue but hopefully it will help us diagnose it.

src/commands/test/lib/junit-xml-util.ts
```diff
@@ -12,46 +11,26 @@ export class JUnitXmlUtil extends XmlUtil {
...
+        let name: string = "unknown";
+        const matches: RegExpMatchArray = entry.path.match("^(.*)_TEST.*");
+        if (matches && matches.length > 1) {
+          name = matches[1].replace(/\./gi, "_");
+        }
+
+        self.appendToTestNameTransformation(xml, name);
+        self.removeIgnoredTransformation(xml);
+
+        mainXml = self.combine(mainXml, xml);
```
owenniblock: Just noticed this - should this be the same as the NUnit one?

```
if (mainXml) {
          mainXml = self.combine(mainXml, xml);
        } else {
          mainXml = xml;
     }
```

src/commands/test/lib/junit-xml-util.ts
```diff
@@ -12,46 +11,26 @@ export class JUnitXmlUtil extends XmlUtil {
...
+        const xml: Document = new DOMParser(XmlUtil.DOMParserConfig).parseFromString(fs.readFileSync(fullPath, "utf-8"), "text/xml");
+
+        let name: string = "unknown";
+        const matches: RegExpMatchArray = entry.path.match("^(.*)_TEST.*");
+        if (matches && matches.length > 1) {
+          name = matches[1].replace(/\./gi, "_");
+        }
+
+        self.appendToTestNameTransformation(xml, name);
+        self.removeIgnoredTransformation(xml);
```
owenniblock: We should investigate if there is there any reason why we can't add this:`self.removeEmptySuitesTransformation(xml);`? I think if we're smart about it we may be able to refactor this even further so that lines 17 to 28 here can be shared by both JUnit and NUnit classes? (Please correct me it I'm mistaken!)

src/commands/test/lib/junit-xml-util.ts
owenniblock: If we completely merge these sections between NUNit and JUnit - we could use this logic though? Because `mainXml` will only return `false` for JUnit?

src/commands/test/lib/junit-xml-util.ts
AndreyMitsyk: NUnit and JUnit use different xml structure. E.g. `removeIgnoredTransformation` are completely different. I see that we have empty test-suite only inside nunit xml.

owenniblock: That's a shame - could we do something like this? (please excuse the pseudo-code):

```
get the xml document (shared code)
perform regex matching (shared code)
process xml (these `self.appendToTestNameTransformation`, `self.removeIgnoredTransformation` etc) (class specific override of abstract function)
combine xml (shared code)
```

src/commands/test/lib/junit-xml-util.ts
AndreyMitsyk: We have different regex as well and use of `self.removeEmptySuitesTransformation(xml);` for nunit. In addition, implementation and usage of appendToTestNameTransformation and removeIgnoredTransformation is different, so we need multiple parameters for the shared function. 
For me, when number of parameters for some shared code is large - it's a reason to not reuse everything 😃

owenniblock: Of course you're correct! I hadn't noticed the difference in the regex.

PR #458, opened 2018-06-29
Feature/check new version

AndreyMitsyk:
Added version check via update-notifier according to comment here #444 

Verified on Mac (node v9.11.1) and Windows (node v9.10.0) by modifying package.json locally:

image

image


src/util/commandline/command.ts
```diff
@@ -194,4 +195,17 @@ export class Command {
 
     return input;
   }
+
+  private checkForUpdates(): void {
```
christav: Good code, but it doesn't belong here. This check should be performed in bin/appcenter.js instead. That's the main entry point.

It should also listen for the `--quiet` switch and not print anything if the "quiet" switch is turned on. Otherwise, automation scripts will randomly start failing due to extra unexpected output.

AndreyMitsyk: Changed. Please have a look at new iteration.

PR #462, opened 2018-07-26
--test-output-dir is only valid for run {manifest|uitest|appium}

john7doe:


src/commands/test/lib/run-tests-command.ts
```diff
@@ -137,16 +135,7 @@ export abstract class RunTestsCommand extends AppCommand {
...
-          if (this.testOutputDir) {
-
-              // Download json test result
-              const testReport: TestReport = await client.test.getTestReport(testRun.testRunId, this.app.ownerName, this.app.appName);
-              if (testReport.stats.artifacts) {
-                await downloadUtil.downloadArtifacts(this, this.streamingOutput, this.testOutputDir, testRun.testRunId, testReport.stats.artifacts);
-                await this.mergeTestArtifacts();
-              }
-          }
+          await this.afterCompletion(client, testRun);
```
owenniblock: Could we send `this.streamingOutput` here rather than having to make it `protected`? Or have a protected accessor of some kind rather than a variable?

I'm not sure what the normal practice is here with TypeScript (I have my C# head on) so let me know if this is not a problem :)

src/commands/test/lib/run-tests-command.ts
```diff
@@ -80,13 +78,8 @@ export abstract class RunTestsCommand extends AppCommand {
...
   vstsIdVariable: string;
 
-  @help(Messages.TestCloud.Arguments.TestOutputDir)
-  @longName("test-output-dir")
-  @hasArg
-  testOutputDir: string;
-
   protected isAppPathRequired = true;
-  private readonly streamingOutput = new StreamingArrayOutput();
+  protected readonly streamingOutput = new StreamingArrayOutput();
```
owenniblock: See below - could we keep this `private` and access it in some other way?

Issue #456, opened 2018-06-21
Updates released with --outputDir option loses codesign signature


Liampronan:
I'm running into a similar issue that I came across in the codepush-cli repo -- originally described here: https://github.com/Microsoft/code-push/issues/544. To summarize, when I try to build a react-native bundle to a custom folder and then attempt to release the signed folder with `codepush release`, my clients with the associated public key can't download the build; the newly released build is uploaded to codepush, but clients show the `Error! Public key was provided but there is no JWT signature within app bundle to verify` error. I can successfully sign/upload/download builds without custom folder path `release-react`. My use case for the custom build path is that it's the only way I can seem to support OTA assets, e.g., images.

**first step - build bundle to custom folder:** `react-native bundle --platform android --entry-file index.android.js --dev false --bundle-output ./release/android/index.android.bundle --sourcemap-output ./release/android/main.android.bundle.map --assets-dest ./release/android`

**second step - release custom folder using signed key:** `appcenter codepush release -a Digit/Digit-Android -d Staging -r 100 -c ./release/android/ -t 2.25 --private-key-path ./code_push_signing_key.pem --token $CODE_PUSH_CLI_TOKEN`

here's a PR that looks to have fixed it but I am running latest cli version (`app-center -v` yields `appcenter version 1.0.17`) https://github.com/Microsoft/appcenter-cli/pull/346.

when I run `release-react` without a custom folder path, I can successully release with the private-ke-signing: `appcenter codepush release-react -a Digit/Digit-Android -d Staging -r 100 --private-key-path ./code_push_signing_key.pem --entry-file index.android.js --sourcemapOutput ./release/main.android.bundle.map --token $CODE_PUSH_CLI_TOKEN`

Any thoughts on this?

amchew:
Hey @Liampronan, the PR that looks to have fixed the issue is in v1.0.8 and your version v1.0.17 is the latest version. We're investigating this. FYI @Zakeelm

Liampronan:
thanks! it's very possible I'm doing something wrong. for now I am just using the `release-react` cmd and falling back to base64 encoding images in OTA updates

Zakeelm:
Hey there! 

@alexandergoncharov will be looking into this and investigating. He'll keep you posted on the status of this as he learns more about the issue.

alexandergoncharov:
Hi all,

Sorry for delay. Fix still in progress. This fix require implementing additional logic and more time for this.

[...]

PR #466, opened 2018-08-08
Fix release command with private key case

alexandergoncharov:
**Issue**: We lose private key during release for RN from custom path because there should be root folder "CodePush" for React-Native app. (For Cordova it should be "www").  

**Suggestion fix**: In private key enabled case we should check that it is React-Native app and the last folder isn't "CodePush". In this case we should copy all contents to temp folder with root "CodePush" folder and then make release from here.

### DIFF
--- a/src/commands/codepush/lib/file-utils.ts
+++ b/src/commands/codepush/lib/file-utils.ts
@@ -42,7 +42,7 @@ export function moveReleaseFilesInTmpFolder(updateContentsPath: string): string 
 }
 
 export function getLastFolderInPath(path: string): string {
-  const splittedPath = path.split("/").filter((el) => { return el !== ""; });
+  const splittedPath = normalizePath(path).split("/").filter((el) => { return el !== ""; });
   if (isDirectory(path)) {
     return splittedPath[splittedPath.length - 1];
   } else {
--- a/src/commands/codepush/lib/file-utils.ts
+++ b/src/commands/codepush/lib/file-utils.ts
@@ -3,6 +3,7 @@ import * as path from "path";
 import * as os from "os";
 import * as rimraf from "rimraf";
 import * as temp from "temp";
+import * as pfs from "../../../util/misc/promisfied-fs";
 
 export function isBinaryOrZip(path: string): boolean {
   return path.search(/\.zip$/i) !== -1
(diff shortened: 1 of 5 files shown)


owenniblock:
Thanks for contributing @alexandergoncharov - we'll take a look at this as soon as we can.

Would it be possible to add some tests for this functionality?

I don't know the `codepush` command well enough to say if what we're doing here is good or not so I'll seek advise but some initial questions you might be able to answer:
- Is it possible that this change will break the command for existing react-native users? (Would it make sense for this functionality to be on a switch or would it be better if it _does_ break and users have to fix it)
- You mention that for Cordova it should be `www`? Is this already handled by the cli? Or should we add handling for that as part of this PR?

Thanks!

alexandergoncharov:
Hi @owenniblock ,

Yeah, I think that I can add tests for this a bit later. Sorry for delay.

About your questions:
1) No, it shouldn't break any commands/logic for users. This changes just fix path for releasing context to make compatibility with code-push SDK for RN which require for working sign feature. Forgot to mention that this fix is for https://github.com/Microsoft/appcenter-cli/issues/456

2) Yes, that's correct. No, cli doesn't handle this because Cordova automatically build content in `www` folder. And users use this folder for releasing.

Also, I have some improvements for code organisation for my PR. I'll update it today.

Thanks,
Alexander

owenniblock:
Hi @alexandergoncharov - is this ready for a re-review? Or are you still working on it? :)

alexandergoncharov:
Hi @owenniblock ,

Code is ready for review. For now I'm working on tests for this.

alexandergoncharov:
Hi @owenniblock ,

Sorry for delay. I added test for implemented logic. I think it is ready for review. :)

Please let me know if you have any questions.

owenniblock:
Thanks @alexandergoncharov - sorry I've been away. I'll take a look at this as soon as possible.

src/commands/codepush/lib/file-utils.ts
```diff
@@ -27,6 +28,30 @@ export function copyFileToTmpDir(filePath: string): string {
...
+    fs.mkdirSync(tmpUpdateContentsPath);
+
+    if (isDirectory(updateContentsPath)) {
+      await pfs.cp(normalizePath(updateContentsPath), normalizePath(tmpUpdateContentsPath));
+    } else {
+      const targetFileName = path.parse(updateContentsPath).base;
+      await pfs.cpFile(normalizePath(updateContentsPath), path.join(tmpUpdateContentsPath, targetFileName));
+    }
+
+    return Promise.resolve(tmpUpdateContentsPath);
```
max-mironov: nit: no need to return `Promise` from the `async` function, should be wrapped automatically in a promise.

alexandergoncharov: @max-mironov Yeah, That's right. Fixed. Thanks!

PR #468, opened 2018-08-14
Added verification to raise an error when 'include' argument provided for Espresso and XCUITest.

AndreyMitsyk:
**Problem:** `--include` argument is not supported for `Espresso` and `XCUITest`.

**Changes:**
- Added verification to raise an error when 'include' argument provided for Espresso and XCUITest.
- Unified error messages.
- Changed help messages for 'include' argument for Espresso and XCUITest.
- Added tests.

---
Error messages for Espresso/XCUITest:
```
Error: Argument --include cannot be used for Espresso
Error: Argument --include cannot be used for XCUITest
```

Help messages for Espresso/XCUITest/other:
```
--include <arg>           Attribute is not supported for Espresso
--include <arg>           Attribute is not supported for XCUITest
--include <arg>           Additional files and directories to include. The value must be either path relative to the input directory, or be in format "targetDir=sourceDir"
```

### DIFF
--- a/src/commands/test/lib/help-messages.ts
+++ b/src/commands/test/lib/help-messages.ts
@@ -96,6 +96,8 @@ export module Messages {
       export const DownloadTestOutputDir = "Directory to download the .zip file(s) into for merging";
       export const MergedFileName = "Name of the merged XML file";
       export const Continuous = "Continuously checks the test run status until it completes";
+
+      export const NotSupported = "Attribute is not supported";
     }
   }
 }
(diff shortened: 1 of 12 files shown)


src/commands/test/lib/help-messages.ts
```diff
@@ -96,6 +96,8 @@ export module Messages {
       export const DownloadTestOutputDir = "Directory to download the .zip file(s) into for merging";
       export const MergedFileName = "Name of the merged XML file";
       export const Continuous = "Continuously checks the test run status until it completes";
+
```
owenniblock: Remove extra newline here.

PR #470, opened 2018-08-16
UITest-Android - Android tests launched by Build don't use the latest stable version of Xamarin.UITest

AndreyMitsyk:
**Problem:** 
Android tests launched by Build are using UITest 2.2.1 which doesn't have support for the latest Android. As a result customers are getting errors when tests launched by the build.

**Root-cause:** Default UITest templates generated by `test generate uitest --output-path  template-path --platform android` command contains hardcoded reference on UITest 2.2.1.

**Solution:** Update UITest version after downloading template.

**Changes:**

- Added base logic to process test templates after downloading on CLI side.
- Implemented logic to retrieve UITest version via NuGet api.
- Implemented logic to replace uitest version in packages.config file.
- Implemented logic to replace uitest version in *.csproj file.
- Added test data.
- Implemented test to verify version replacement.

---
`test generate uitest` command verified locally for Android/iOS with different parameters. If NuGet api returns an error (e.g. due to connection issue), the default template will be used.

Regex for `packages.config` replaces version here:
```
<package id="Xamarin.UITest" version="2.2.1" targetFramework="net45" />
```

Regex for `AppCenter.UITest.Android.csproj` replaces version here:
```
<Reference Include="Xamarin.UITest">
  <HintPath>..\packages\Xamarin.UITest.2.2.1\lib\net45\Xamarin.UITest.dll</HintPath>
</Reference>
```


owenniblock:
I'm hesitant to approve this because it feels like a hack - I'm trying to decide:
- Why we're including the project files in S3 in the first place? Would it make more sense to add them to the cli project in some way? They are (as far as I'm aware) the only consumer of these files.
- If there's any way to increase the UITest version without manually editing and updating the file here?

So an idea for discussion:
Instead of downloading and updating the project files - we include them in appcenter-cli - then we manually update the UITest version (for now) to 2.2.5.

We add some automated way of updating the value of UITest version in the files we've added to appcenter-cli (maybe on UITest release definition?) which automatically creates a PR / commit / something.

That way if we have breaking changes to UITest - we can catch them before the cli is approved and released. Also in future we can add integration tests to this area in appcenter-cli.

AndreyMitsyk:
@owenniblock 
I assume that original idea with s3 was to update templates without updating CLI.
For version updates we can use nuget cli `nuget update AppCenter.UITest.Android.sln -Id Xamarin.UITest` but in this case we introducing new dependency which is bad.
I can include templates in CLI with placeholder instead of version and replace it during `generate` command.
The core problem with solution "which automatically creates a PR" is the users needs new CLI to use latest UITest version.

owenniblock:
@AndreyMitsyk 
> I assume that original idea with s3 was to update templates without updating CLI

I'm pretty sure that was the original idea - but with the version number in the .zip file that becomes very confusing! You'd have to release a zip with the same version (or release the cli with an updated reference to the zip file). I don't think this process has ever been quite right.

>  users needs new CLI to use latest UITest version

I'm not entirely sure this is a bad thing - if we introduce breaking changes in UITest we wouldn't want the UITest release to break the launch tests in the CLI - instead we should be able to fix any breaks before releasing the cli.

AndreyMitsyk:
@owenniblock 
The original bug here that old UITest does not work with new devices after executing tests via build. Without auto-updating we can't solve this quickly. With updates in source code it's possible that we will have minor CLI updates with UITest version only.

I'm just uncomfortable that we have dependency between CLI and Xamarin.UITest. Right now we faced with this issue on Android, but I guess that in we will have same problem with iOS 12 soon: we will have iOS 12 devices, but default UITest version in CLI can not be used with latest devices.

AndreyMitsyk:
My suggestion is:
1) Move templates to CLI repo
2) Remove old download zip logic
3) Replace version during `generate` command.
I can add parameter to use update logic only if some parameter is here, e.g.:
`test generate uitest --output-path template-path --platform android --latest`
We can use this parameter when we are trying to call `generate` from build.

owenniblock:
Let's do 1 & 2.

3 would require a change (and release) on App Center Build - we can always introduce that switch in retrospect later if we need to.

PR #472, opened 2018-08-20
Cleanup tests.

AndreyMitsyk:
- Removed excessive step from appium tests.
- Changed test output dir for generate uitest command.


owenniblock:
Just checking out why the build failed before I merge.

PR #473, opened 2018-08-20
Updates npm packages

owenniblock:
### Motivation
We need to keep the npm packages up to date - they've not been updated for a while so this attempts to redress the balance.

### TODO:
- [x] Build & test runs successfully
- [x] Smoke test on compiled code
- [x] Discuss whether we should test thoroughly before release or risk breaks & then fix


owenniblock:
@john7doe - we would appreciate a 3rd pair of eyes on this (if you have time) - will discuss tomorrow and take you off Reviewers if you've got too much on :)

src/commands/crashes/lib/mdfind.ts
```diff
@@ -0,0 +1,103 @@
+/**
+ * Node module for searching OSX Spotlight, using the built-in mdfind shell command.
+ * Forked from brandonhorst/node-mdfind github repo.
```
john7doe: add reason to the copy (insecure deps if I remember correctly)

AndreyMitsyk: done.

PR #479, opened 2018-08-27
Improved error handling for generate uitest command.

AndreyMitsyk:
Improved error handling for generate uitest command:
- Moved code to copy templates to separate method.
- Added separate `catch` blocks for retrieving and replacing version.
- Added copying of original templates if replace operation failed.

### DIFF
--- a/src/commands/test/generate/uitest.ts
+++ b/src/commands/test/generate/uitest.ts
@@ -20,16 +20,25 @@ export default class GenerateUITestCommand extends GenerateCommand {
     const packageFilePath = path.join(this.outputPath, `AppCenter.UITest.${platform}/packages.config`);
     const projectFilePath = path.join(this.outputPath, `AppCenter.UITest.${platform}/AppCenter.UITest.${platform}.csproj`);
 
-    try {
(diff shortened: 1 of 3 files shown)


src/commands/test/generate/uitest.ts
```diff
@@ -20,16 +20,25 @@ export default class GenerateUITestCommand extends GenerateCommand {
...
 
+    try {
       // Replace version inside packages.config file
       await this.replaceVersionInFile(packageFilePath, /(id="Xamarin\.UITest" version=")(\d+(\.\d+)+)/, latestVersion);
 
       // Replace version inside *.csproj file
       await this.replaceVersionInFile(projectFilePath, /(packages\\Xamarin\.UITest\.)(\d+(\.\d+)+)/, latestVersion);
     } catch (e) {
-      console.warn("Can't update UITest version. Using default version from template. Details: " + e);
+      await this.copyTemplates();
```
owenniblock: Is it possible to add a few tests for this functionality? So we can verify that the `copyTemplates` function is called when `replaceVersionInFile` throws an error?

AndreyMitsyk: Added two tests 😄

PR #492, opened 2018-10-05
Add additional info(Metadata and Metrics) for codepush deployment list

alexandergoncharov:
**Issue:** `code-push` cli shows metadata and metrics info about deployment with `deployment list command`.
`appcenter-cli` shows just keys without any additional info:
image

**Solution:**
Added `-k` option(`--displayKeys` is full name) for showing keys.
Added Metadata and Metrics for deployments.

Without `-k` option:
image

With `-k` option: 
image


src/commands/codepush/deployment/list.ts
```diff
@@ -22,7 +28,17 @@ export default class CodePushDeploymentListListCommand extends AppCommand {
       const httpRequest = await out.progress("Getting CodePush deployments...", clientRequest<models.Deployment[]>(
         (cb) => client.codePushDeployments.list(app.ownerName, app.appName, cb)));
       deployments = httpRequest.result;
-      out.table(out.getCommandOutputTableOptions(["Name", "Key"]), deployments.map((deployment) => [deployment.name, deployment.key]));
+
+      if (this.displayKeys) {
```
owenniblock: If we're defaulting this to `false` and changing the default behaviour of the command - can we add some help text to the default output to guide users who may already be using the `Name, Key` table output?

That way users can quickly see what they need to add to the command to see the output they were using previously.

src/commands/codepush/deployment/list.ts
```diff
@@ -34,4 +50,76 @@ export default class CodePushDeploymentListListCommand extends AppCommand {
...
+      }
+
+      return [deployment.name, metadataString, metricsString];
+    }));
+  }
+
+  private generateMetricsString(releaseMetrics: models.CodePushReleaseMetric, releasesTotalActive: number): string {
+    let metricsString: string = "";
+
+    const activePercent: number = (releaseMetrics.downloaded) ? releaseMetrics.active / releasesTotalActive * 100 : 0.0;
```
owenniblock: Although it looks like there are no ways for `releasesTotalActive` to be `0` here - could we add protective code in case the calling code changes (so we don't get a divide by zero error here).

alexandergoncharov: Yeah, you are absolutely right. Thanks. Fixed.

src/commands/codepush/deployment/list.ts
alexandergoncharov: Yeah, It would be better. Added text before table. Thanks

alexandergoncharov:
Sure, I'll add tests for this functionality and ping you :)

PR #500, opened 2018-11-16
make app option optional for test wizard

john7doe:


src/commands/test/wizard.ts
```diff
@@ -34,21 +35,35 @@ export default class WizardTestCommand extends AppCommand {
...
   private isAndroidApp: boolean;
+  private _selectedApp: DefaultApp = null;
 
   constructor(args: CommandArgs) {
     super(args);
     this._args = args;
   }
 
+  private async selectedApp(client: AppCenterClient): Promise<DefaultApp> {
+    if (this._selectedApp == null) {
```
AndreyMitsyk: Should it handle `undefined` value? If so, you can just use `if (!this._selectedApp) {`

PR #509, opened 2018-11-26
Fixes version of event-stream to 3.3.4 due to compromised release 3.3.6

owenniblock:
### Motivation
A pre-emptive change due to a compromised version of `event-stream@3.3.6` being released. Whilst this vulnerability will not cause an issue on `appcenter-cli` (it's targeting packages which use `ps-tree`) we're fixing the version to `3.3.4` because:
- Confidence in this module has been damaged - we should assess each new version or switch to a different library
- We need to clear out any cached instances of the offending version

### Other notes
- [More information about the event-stream vulnerability can be found here](https://github.com/dominictarr/event-stream/issues/116)

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -2031,18 +2031,27 @@
       }
     },
     "event-stream": {
-      "version": "3.3.6",
-      "resolved": "https://registry.npmjs.org/event-stream/-/event-stream-3.3.6.tgz",
-      "integrity": "sha512-dGXNg4F/FgVzlApjzItL+7naHutA3fDqbV/zAZqDDlXTjiMnQmZKu+prImWKszeBM5UQeGvAl3u1wBiKeDh61g==",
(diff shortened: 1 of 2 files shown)


owenniblock:
@john7doe review request is an FYI - I think I'm going to go ahead and merge and release this.

phated:
This should definitely be a temporary fix and you shouldn't be relying on software the author hasn't maintained in years.  Also streams 1 are super old.

owenniblock:
Thanks for the input @phated - I agree. I've suggested we do an audit of all of our dependencies to highlight which ones are no longer maintained and look at switching to different libraries or helping maintain the OS libraries if they're popular / it makes sense to do so etc.

I don't think we'll have a quick turnaround on this so if you want to contribute to the project and switch `event-stream` for something else I'm happy to discuss your favoured approach here!

PR #512, opened 2018-11-30
Sample of removing event-stream for review

phated:
@owenniblock as per https://github.com/Microsoft/appcenter-cli/pull/509#issuecomment-442772582 - I wanted to put together an example of how to remove event-stream and see what you thought.  Both `through2` and `split2` are maintained by node core contributors and have minimal dependencies (both have readable-stream, maintained by node core. And through2 has xtend which @raynos has long frozen activity on)

Please give me feedback and we can proceed from there.


owenniblock:
@phated - thanks for contributing! It looks good to me - I'll check those two libraries and assuming it all looks fine I test here and then merge this in (probably some time next week I would guess).

It looks like a minor dependency issue on the CI server:

```
npm ERR! cipm can only install packages when your package.json and package-lock.json or npm-shrinkwrap.json are in sync. Please update your lock file with `npm install` before continuing.
npm ERR! 
npm ERR! 
npm ERR! Missing: split2@^3.0.0
npm ERR! Invalid: lock file's through2@2.0.3 does not satisfy through2@^3.0.0
npm ERR! Missing: @types/split2@^2.1.6
npm ERR! Missing: @types/through2@^2.0.34
npm ERR! 
```

Please could you take a look at fixing that before we test over here?

Thanks again 🙂

amchew:
Shout out to @phated for contributing!! 💯 Your contributions help us make the code better, so thank you again!

owenniblock:
I'm a bit worried that `split2` hasn't been touched for a little while (it's a minor grumble though) - otherwise those module choices look good to me.

phated:
Split2 is by Matteo, who is the Principal Architect at nearform and is like the head of the node streams working group. Libraries don't need to be constantly updated to be maintained. Changes got us into the event-stream mess, after all.

owenniblock:
@phated - please don't take it as a dig at either you, Matteo Collina or the library itself - it was just an observation I thought worth writing down.

Whilst changes caused the event-stream mess - the underlying catalyst was a project whose maintainer had let the project lapse who then passed maintenance on to an unvetted 3rd party.

I looked at the projects and decided that this _wasn't_ a worry here but wanted to make a note of it (in the unlikely event that we find issues - we may decide to contribute if we need to - so it felt like a worthy note to make).

phated:
I'll update the package-lock but did you know that relying on a package-lock in a module published to npm actually puts your users at risk?  npm does not cascade package-lock files upon npm install so when you include one in your module, your dev and CI environments will likely be using different versions of packages than your end-users are installing.  If you want cascading lockfiles, you need to use a shrinkwrap file, not a package-lock (package-lock files are meant to be used in end-user application code).

phated:
I've done more removal and cleanup to get closer to removing event-stream.  The one hang up I am having is that the equivalent package to `es.writeArray` is `concat-stream` but the type definitions at `@types/concat-stream` hard code the callback data as Buffer when they actually can be a bunch of different types.

I thought maybe I could do custom types but I have no idea how to do that in the project.

I also think that the `.pipe` calls on every stream should be changed to use https://nodejs.org/api/stream.html#stream_stream_pipeline_streams_callback but that would require pulling in the `readable-stream` module to have support in node <10.  The problem I encountered there was that there is no `@types/readable-stream` but it would be just a copy of the node-core `stream` types.

phated:
ping

owenniblock:
Hi @phated - sorry for the delay. I'm deep in a piece of feature work at the moment and haven't had time to check this out. It'll probably be next week before I get to it.

@amchew - please discuss with Mads if you'd like to get this prioritised so we can find a way to fit it in sooner.

phated:
It's extremely disappointing that my last changes were made 11 days ago and this is still sitting 😭

amchew:
Hey @phated ! This is on me, I dropped the ball on this one. I will work on this today, and get back. Sorry for the delay!

phated:
@owenniblock can do! Should I comment them, skip them or delete them (or some variation)?  Thanks for pushing forward on this!

owenniblock:
@phated - please delete them. I'll refer to the diff in the bug I raise in Azure DevOps so we don't forget.

phated:
I believe I've finished removing it but I'm having a little trouble running the tests locally.  Pushed it to check on CI.

phated:
@owenniblock awesome! Thanks for getting this merged 😸

Issue #504, opened 2018-11-20
CodePush SourceMap-Output doesn't work with Output-Dir


TheCodeKing:
I'm trying to generate a CodePush bundle for a React-Native project, and write out the source maps to a folder outside of the CodePush location. I'm using version 1.1.7.

When I specify the source maps location, it's completely ignored and maps get bundled into the CodePush package. Is there ever a scenario where you would distribute source maps to users?

```
appcenter codepush release-react -a "test/Test-App" 
  --deployment-name Staging 
  --entry-file index.js 
  --output-dir "/tmp/output" 
  --sourcemap-output "/tmp/maps/main.jsbundle.map"
```

If I remove the `--output-dir` param then it works, but then I can't set the bundle output location which I also need.

```
appcenter codepush release-react -a "test/Test-App" 
  --deployment-name Staging 
  --entry-file index.js 
  --sourcemap-output "/tmp/maps/main.jsbundle.map"
```

If I try and specify the `--output-dir` without `--sourcemap-output`, then the source maps still get generated.

```
appcenter codepush release-react -a "test/Test-App" 
  --deployment-name Staging 
  --entry-file index.js 
  --output-dir "/tmp/output" 
```

TheCodeKing:
Problem seems to be this line

https://github.com/Microsoft/appcenter-cli/blob/7759122a660276834f2e8a6203893c0942f8623b/src/commands/codepush/release-react.ts#L131

Should be:

```
if (this.outputDir && !this. sourcemapOutput) {
```

amchew:
Thanks @TheCodeKing for flagging this to us! We'll investigate and get back shortly.

alexandergoncharov:
Hi @TheCodeKing, @amchew ,

@TheCodeKing , Good catch! I removed this `if` block to prevent forced generating sourcemaps and prepared PR: https://github.com/Microsoft/appcenter-cli/pull/516
Could you please review it and let me know if you have any questions or suggestions.

Thanks,
Alexander

amchew:
Hi @TheCodeKing - could you please review and let us know if you have any questions? If not, let's proceed to merge it on Monday end of day 12/17 @alexandergoncharov.

TheCodeKing:
@alexandergoncharov looks all good from my side. Thanks for the fix :)

amchew:
[...]

PR #516, opened 2018-12-05
Fix generating sourceMaps for react-native CodePush

alexandergoncharov:
**Issue:** In case if user uses `--outputDir` option(custom directory for generating bundle) SourceMap will be generated in this `outputDir` directory even if user provide custom path for sourceMaps with `--sourcemap-output` option.
Note: SourceMaps is not generated in default case(without `--outputDir` and `--sourcemap-output` options)

**Solution:** SourceMap should be generated only if user provide some path for this in  `--sourcemap-output` option. 
Removed `if` block for forced generating sourceMaps.

Related issue: https://github.com/Microsoft/appcenter-cli/issues/504

### DIFF
--- a/src/commands/codepush/release-react.ts
+++ b/src/commands/codepush/release-react.ts
@@ -128,10 +128,6 @@ export default class CodePushReleaseReactCommand extends CodePushReleaseCommandS
       }
     }
 
-    if (this.outputDir) {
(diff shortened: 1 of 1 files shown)


kufufu9:
Sourcemap will be generated in `outputDir` and uploaded to codepush server when user declare `--outputDir` option. But I needn't the sourcemap.

owenniblock:
Sorry for the delay - merging this! Thanks for your contribution :)

PR #517, opened 2018-12-13
Added support content

jwargo:
There was no reference to our support options.


owenniblock:
Lovely, thanks @jwargo !

PR #522, opened 2019-01-11
Add option to specify the file name when downloading a build

lucasbento:
This PR adds a way of specifying the file name (by `--file` or `-f` option) for the `download` build command.

My use-case is that I need to download the build on a CI and I want to download it with the same file name every time.

### DIFF
--- a/src/commands/build/download.ts
+++ b/src/commands/build/download.ts
@@ -45,6 +45,12 @@ export default class DownloadBuildStatusCommand extends AppCommand {
   @hasArg
   public directory: string;
 
+  @help("Download destination file")
+  @shortName("f")
+  @longName("file")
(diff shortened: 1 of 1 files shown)


lucasbento:
@owenniblock: cool, thanks for approving.

If you have any suggestions for the help text I can send another PR. :)

owenniblock:
I think we should change both this help text and the directory one to make them a little more friendly :) something like:

> "Directory to download the file into"

> "Name of the downloaded file. Optional parameter to override the default auto-generated file name"

I'm going to release today and it's such a minor thing - it's totally up to you if you want to send it over :)

lucasbento:
@owenniblock: submitted a PR for it #524. :)

Issue #527, opened 2019-01-17
I cannot run any command


ahryne:
I using windows 10, with nvm for windows. (node 8.15.0 / 10.15.0)
If i run related command with my account, returned error.

~~~
appcenter apps list
| Getting app list ...events.js:183
      throw er; // Unhandled 'error' event
      ^

TypeError: Invalid non-string/buffer chunk
    at validChunk (C:\Users\seonguk.cho\AppData\Roaming\npm\node_modules\appcenter-cli\node_modules\through2\node_modules\readable-stream\lib\_stream_writable.js:304:10)
    at DestroyableTransform.Writable.write (C:\Users\seonguk.cho\AppData\Roaming\npm\node_modules\appcenter-cli\node_modules\through2\node_modules\readable-stream\lib\_stream_writable.js:332:62)
    at Pumpify.ondata (C:\Users\seonguk.cho\AppData\Roaming\npm\node_modules\appcenter-cli\node_modules\readable-stream\lib\_stream_readable.js:531:20)
    at emitOne (events.js:116:13)
    at Pumpify.emit (events.js:211:7)
    at readableAddChunk (C:\Users\seonguk.cho\AppData\Roaming\npm\node_modules\appcenter-cli\node_modules\readable-stream\lib\_stream_readable.js:198:18)
    at Pumpify.Readable.push (C:\Users\seonguk.cho\AppData\Roaming\npm\node_modules\appcenter-cli\node_modules\readable-stream\lib\_stream_readable.js:157:10)
    at Pumpify.Duplexify._forward (C:\Users\seonguk.cho\AppData\Roaming\npm\node_modules\appcenter-cli\node_modules\duplexify\index.js:170:26)
    at WinCredStoreParsingStream.onreadable (C:\Users\seonguk.cho\AppData\Roaming\npm\node_modules\appcenter-cli\node_modules\duplexify\index.js:134:10)
    at emitNone (events.js:106:13)
~~~

~~~
appcenter login
| Logging out current user...events.js:183
      throw er; // Unhandled 'error' event
      ^

TypeError: Invalid non-string/buffer chunk
    at validChunk (C:\Users\seonguk.cho\AppData\Roaming\npm\node_modules\appcenter-cli\node_modules\through2\node_modules\readable-stream\lib\_stream_writable.js:304:10)
[...]

PR #528, opened 2019-01-17
Fixed win-token-store logic.

AndreyMitsyk:
https://github.com/Microsoft/appcenter-cli/issues/527
According to existing `win-credstore-parser` logic, `through2` should be used in `objectMode` on windows. Similar logic for osx looks correct.

@owenniblock please have a look at this PR. The current version is broken for windows, so new release required after merging this PR.


AndreyMitsyk:
Test on windows are ok:
image

PR #535, opened 2019-01-24
Support breakpad symbol uploading

kfcampbell:
This change adds support for uploading Breakpad symbols via the `crashes upload-symbols` command. Previously, all types of symbols were uploaded with the symbol_type attribute set to "Apple". 

Now, Breakpad symbols are able to be passed in using the `-b` flag.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "appcenter-cli",
-  "version": "1.1.12",
+  "version": "1.1.11",
   "description": "Command line tool for Visual Studio App Center",
   "keywords": [
     "microsoft",
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "appcenter-cli",
-  "version": "1.1.11",
+  "version": "1.1.12",
   "description": "Command line tool for Visual Studio App Center",
   "keywords": [
     "microsoft",
--- a/src/commands/crashes/upload-symbols.ts
+++ b/src/commands/crashes/upload-symbols.ts
@@ -48,7 +48,7 @@ export default class UploadSymbols extends AppCommand {
   @hasArg
   public sourceMapPath: string;
 
-  @help("Path to a breakpad symbols zip collection.")
+  @help("Path to a zip file containing Breakpad symbols.")
   @shortName("b")
   @longName("breakpad")
   @hasArg
@@ -81,7 +81,7 @@ export default class UploadSymbols extends AppCommand {
     }
 
     let pathToZipToUpload: string;
-    if (typeof (zip) === "string") {
+    if (typeof(zip) === "string") {
       // path to zip can be passed as it is
       pathToZipToUpload = zip;
     } else {
@@ -179,8 +179,6 @@ export default class UploadSymbols extends AppCommand {
       throw failure(ErrorCodes.InvalidParameter, "'--symbol' and '--breakpad' switches are mutually exclusive");
     } else if (!_.isNil(this.xcarchivePath) && !_.isNil(this.breakpadPath)) {
       throw failure(ErrorCodes.InvalidParameter, "'--xcarchive' and '--breakpad' switches are mutually exclusive");
-    } else if (!_.isNil(this.breakpadPath) && !_.isNil(this.sourceMapPath)) {
-      throw failure(ErrorCodes.InvalidParameter, "'--breakpad' and '--sourcemap' switches are mutually exclusive");
     }
   }
 
@@ -201,7 +199,7 @@ export default class UploadSymbols extends AppCommand {
     }
 
     let zipToChange: JsZip;
-    if (typeof (zip) === "string") {
+    if (typeof(zip) === "string") {
       // loading ZIP to add file to it
       debug("Loading ZIP into the memory to add files");
       try {
--- a/src/commands/crashes/upload-symbols.ts
+++ b/src/commands/crashes/upload-symbols.ts
@@ -8,6 +8,7 @@ import { DefaultApp } from "../../util/profile";
 import UploadSymbolsHelper from "./lib/symbols-uploading-helper";
 import { getSymbolsZipFromXcarchive, packDsymParentFolderContents, getChildrenDsymFolderPaths } from "./lib/subfolder-symbols-helper";
 import { createTempFileFromZip } from "./lib/temp-zip-file-helper";
+import { SymbolType } from "./lib/symbols-uploading-helper";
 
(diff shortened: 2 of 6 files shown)


package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "appcenter-cli",
-  "version": "1.1.11",
+  "version": "1.1.12",
```
owenniblock: You don't need to increate the version (we'll do that when we release so that we can roll up any other changes in a single version number). Please can you revert this change?

src/commands/crashes/upload-symbols.ts
```diff
@@ -47,18 +48,30 @@ export default class UploadSymbols extends AppCommand {
   @hasArg
   public sourceMapPath: string;
 
+  @help("Path to a breakpad symbols zip collection.")
```
owenniblock: Will this be a collection of zip files? Or a collection of other files that get zipped? (Apologies - I don't have the context to know this so wanted to check that this help text would be clear to users :) )

src/commands/crashes/upload-symbols.ts
```diff
@@ -68,7 +81,7 @@ export default class UploadSymbols extends AppCommand {
     }
 
     let pathToZipToUpload: string;
-    if (typeof(zip) === "string") {
+    if (typeof (zip) === "string") {
```
owenniblock: Not a biggie but please try and avoid whitespace changes if possible?

src/commands/crashes/upload-symbols.ts
```diff
@@ -158,10 +171,16 @@ export default class UploadSymbols extends AppCommand {
...
+    if (_.isNil(this.symbolsPath) && _.isNil(this.xcarchivePath) && _.isNil(this.breakpadPath)) {
+      throw failure(ErrorCodes.InvalidParameter, "specify either '--symbol', '--xcarchive', or '--breakpad' switch");
     } else if (!_.isNil(this.symbolsPath) && !_.isNil(this.xcarchivePath)) {
       throw failure(ErrorCodes.InvalidParameter, "'--symbol' and '--xcarchive' switches are mutually exclusive");
+    } else if (!_.isNil(this.symbolsPath) && !_.isNil(this.breakpadPath)) {
+      throw failure(ErrorCodes.InvalidParameter, "'--symbol' and '--breakpad' switches are mutually exclusive");
+    } else if (!_.isNil(this.xcarchivePath) && !_.isNil(this.breakpadPath)) {
+      throw failure(ErrorCodes.InvalidParameter, "'--xcarchive' and '--breakpad' switches are mutually exclusive");
+    } else if (!_.isNil(this.breakpadPath) && !_.isNil(this.sourceMapPath)) {
+      throw failure(ErrorCodes.InvalidParameter, "'--breakpad' and '--sourcemap' switches are mutually exclusive");
```
owenniblock: Checking this is true (since `--sourcemap` is not mutually exclusive to any of the other parameters).

src/commands/crashes/upload-symbols.ts
```diff
@@ -182,7 +201,7 @@ export default class UploadSymbols extends AppCommand {
     }
 
     let zipToChange: JsZip;
-    if (typeof(zip) === "string") {
+    if (typeof (zip) === "string") {
```
owenniblock: See whitespace comment above :)

test/commands/crashes/upload-symbols/upload-symbols-test.ts
```diff
@@ -69,10 +69,28 @@ describe("upload-symbols command", () => {
     abortSymbolUploadSpy = Sinon.spy();
   });
 
-  describe("when network requests are successful", () => {
+  describe("when Breakpad network requests are successfu", () => {
```
owenniblock: Small typo in `describe` text (missing `l` off the end)

src/commands/crashes/upload-symbols.ts
kfcampbell: done!

package.json
kfcampbell: done!

src/commands/crashes/upload-symbols.ts
kfcampbell: yeah no problem! this was a remnant of doing an auto-format for this file. i've reverted this and the one below.

src/commands/crashes/upload-symbols.ts
kfcampbell: (see my comment above)

test/commands/crashes/upload-symbols/upload-symbols-test.ts
kfcampbell: good catch! i don't know how that snuck in there.

src/commands/crashes/upload-symbols.ts
kfcampbell: i've changed the wording to "Path to a zip file containing Breakpad symbols." if that's not clear or you'd like me to use different wording, just let me know!

owenniblock: Ah, I think what was confusing me here was the naming `prepareZipFromSymbols` - whereas in this case really we're just sending it a zip file and it's returning the path we send it (which seems a little odd!)

when it's a breakpad - will it always be a .zip? If so - this help text is fine :)

src/commands/crashes/upload-symbols.ts
owenniblock: Sorry for not being clearer.

This is what I meant - but I wasn't sure if the check that breakpad and sourcemap are both being passed was correct (and intentional).

src/commands/crashes/upload-symbols.ts
kfcampbell: Yep! Breakpad symbols will always be zipped, as mentioned in our documentation here: https://docs.microsoft.com/en-us/appcenter/diagnostics/android-ndk

src/commands/crashes/upload-symbols.ts
kfcampbell: Oh I'm sorry, I misunderstood -- that check is actually required. Sourcemaps (as currently implemented in the CLI) only pertain to iOS. Android sourcemaps (to the best of my understanding) would be implemented as "AndroidProGuard". Thus, this check is necessary to make sure the user isn't trying to upload iOS sourcemaps and Android symbols for the same app. I've added it back and resolved this conversation.

If you're think we're good to merge, I think we're good to merge! Just let me know. Otherwise, I'd be happy to discuss this further!

PR #544, opened 2019-02-22
update launch test

john7doe:


src/commands/test/lib/templates/appium/ios/pom.xml
```diff
@@ -4,25 +4,32 @@
...
-      <version>4.1.2</version>
+      <version>6.1.0</version>
     </dependency>
     <dependency>
       <groupId>junit</groupId>
       <artifactId>junit</artifactId>
       <version>4.12</version>
     </dependency>
-    <!-- See https://github.com/xamarinhq/test-cloud-appium-java-extensions -->
+    <!-- See https://docs.microsoft.com/da-dk/appcenter/test-cloud/preparing-for-upload/appium -->
```
owenniblock: This should be `https://docs.microsoft.com/appcenter/test-cloud/preparing-for-upload/appium` so that the user's location determines the iso code?

PR #550, opened 2019-03-14
Update CodePush Release Electron to output sourcemap to a directory

prashantpatali:
This change includes the following -
1. Make CodePush Release Electron consistent with CodePush Release ReactNative by using the same switch for SourceMap as well as providing a new option for SourceMapOutputDir.
2. Make WebPack-CLI as Dependency, so its packaged in the node_modules for use by CodePush Release Electron option.
3. Fixed a NPM Audit issue for gulp-typescript.


src/commands/codepush/release-electron.ts
```diff
@@ -37,11 +37,16 @@ export default class CodePushReleaseElectronCommand extends CodePushReleaseComma
   @hasArg
   public entryFile: string;
 
-  @help("Sourcemap filename of the resulting bundle. If omitted, a sourcemap will not be generated")
+  @help("Path to where the sourcemap for the resulting bundle should be written. If omitted, a sourcemap will not be generated")
```
owenniblock: I think this wording could be confused with the `sourcemap-output-dir` parameter (it's not clear that this defines the name of the file rather than the directory).

I suggest:
> Filename (including path) for the sourcemap of the resulting bundle. If \"sourcemap-output\" and \"sourcemap-output-dir\" are omitted, a sourcemap will not be generated

Is there a particular impetus behind the `longName` change? I wonder if we should include a depreciated `sourcemapFilename` which gets mapped to `sourcemapOutput` with a warning?

src/commands/codepush/release-electron.ts
```diff
@@ -37,11 +37,16 @@ export default class CodePushReleaseElectronCommand extends CodePushReleaseComma
...
   @shortName("s")
-  @longName("sourcemap-filename")
+  @longName("sourcemap-output")
   @hasArg
-  public sourcemapFileName: string;
+  public sourcemapOutput: string;
+
+  @help("Path to folder where the sourcemap for the resulting bundle should be written. Name of sourcemap file will be generated automatically. This argument will be ignored if \"sourcemap-output\" argument is provided. If omitted, a sourcemap will not be generated")
```
owenniblock: Suggested wording:

> Path to directory where the sourcemap for the resulting bundle should be written. Name of sourcemap file will be generated automatically. This argument will be ignored if \"sourcemap-output\" argument is provided. If \"sourcemap-output\" and \"sourcemap-output-dir\" are omitted, a sourcemap will not be generated

owenniblock:
Please could you also remove the `gulp-typescript` update - as this is a dev dependency, we'd rather keep it separate from this PR and handle it internally.

owenniblock:
Perfect, thanks! Merging - I'll not get chance to release immediately - I'll discuss with my manager when we can do the release and let you know :)

PR #554, opened 2019-03-15
Show deprecation warning for -g / --group parameter when distributing

cainejette:
image


pfleidi:
@cainejette After talking to our PMs, we decided to just change the APIs used in this command to something that isn't deprecated. Therefore, we don't have to deprecate this command for now.

More info can be found here: https://github.com/pfleidi/app-center-cli/pull/3

owenniblock:
Thanks @pfleidi - so if I'm reading that correctly we should close this PR and create a new PR from your fork?

pfleidi:
@owenniblock You can close this PR. We'll be opening a new one based on the changes in my fork. Makes sense?

PR #556, opened 2019-03-18
Add `distribute releases edit` command

pfleidi:
This PR will add the `distribute releases edit` command which allows to enable or disable a release.

```
appcenter-cli (edit-release) » ./bin/appcenter.js distribute releases edit --help                                                                                                                            ~/work/microsoft/appcenter-cli

Enables or disables specified release

Usage: appcenter distribute releases edit -r|--release-id <arg> [-a|--app <arg>] <State>

Options:
    -r|--release-id <arg>             Release ID
    -a|--app <arg>                    Specify app in the <ownerName>/<appName> format
    State                             Release state
```

### DIFF
--- a/src/commands/distribute/releases/edit.ts
+++ b/src/commands/distribute/releases/edit.ts
@@ -1,0 +1,71 @@
+import { AppCommand, CommandResult, ErrorCodes, failure, help, success, shortName, longName, required, hasArg, position, name } from "../../../util/commandline";
+import { AppCenterClient, clientRequest, models } from "../../../util/apis";
+import { out } from "../../../util/interaction";
+import { inspect } from "util";
+
+const debug = require("debug")("appcenter-cli:commands:distribute:releases:delete");
+
+@help("Enables or disables specified release")
+export default class EditReleaseCommand extends AppCommand {
+  @help("Release ID")
+  @shortName("r")
+  @longName("release-id")
+  @required
+  @hasArg
+  public releaseId: string;
+
+  @help("Release state")
+  @name("State")
+  @position(0)
+  @required
+  public state: string;
+
+  public async run(client: AppCenterClient): Promise<CommandResult> {
+    const app = this.app;
+
+    const releaseId = Number(this.releaseId);
+    if (!Number.isSafeInteger(releaseId) || releaseId <= 0) {
+      return failure(ErrorCodes.InvalidParameter, `${this.releaseId} is not a valid release id`);
+    }
+
+    const state = this.state;
+    if (state !== "enabled" && state !== "disabled") {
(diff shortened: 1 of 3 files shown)


src/commands/distribute/releases/edit.ts
```diff
@@ -0,0 +1,71 @@
...
+@help("Enables or disables specified release")
+export default class EditReleaseCommand extends AppCommand {
+  @help("Release ID")
+  @shortName("r")
+  @longName("release-id")
+  @required
+  @hasArg
+  public releaseId: string;
+
+  @help("Release state")
```
owenniblock: I think the help text should display the valid states.

src/commands/distribute/releases/edit.ts
```diff
@@ -0,0 +1,71 @@
...
+    }
+
+    let releaseDetails: models.ReleaseDetailsResponse;
+    try {
+      debug("Loading release details");
+      const httpRequest = await out.progress("Loading release details...", clientRequest<models.ReleaseDetailsResponse>(
+        (cb) => client.releases.getLatestByUser(this.releaseId, app.ownerName, app.appName, cb)
+      ));
+      if (httpRequest.response.statusCode >= 400) {
+        throw httpRequest.response.statusCode;
```
owenniblock: I'm not super keen on throwing a number like this - I'd rather handle the error cases here directly and only handle legitimate exceptions in the `catch` block.

test/commands/distribute/releases/edit-test.ts
```diff
@@ -0,0 +1,218 @@
...
+      const result = await command.execute();
+
+      // Assert
+      expect((result as CommandFailedResult).errorMessage).to.eql(expectedErrorMessage);
+    });
+
+    it("if state option wasn't specified", async () => {
+      // Arrange
+      let errorMessage: string;
+      const expectedErrorMessage = "Missing required positional argument State";
```
owenniblock: Not your doing but a note to self - I really don't like the wording here. Using the terminology `required positional argument` seems a bit hard to understand to me.

src/commands/distribute/releases/edit.ts
owenniblock: I see that we do this in other places too - still don't like it but if it's standard then don't worry about this one.

pfleidi: Yeah, I don't like it too much, either but I think this pattern was copied from another command which was using it.

owenniblock: Yeah, I noticed that after I submitted the review 😆 Let's worry about this another day!

hagould: we resolved it for this task! fair feedback

PR #557, opened 2019-03-18
Upgrade distribute api calls

pfleidi:
This PR updates the existing `appcenter distribute release` command to all our new, non-deprecated APIs to do its job.

### DIFF
--- a/src/commands/distribute/release.ts
+++ b/src/commands/distribute/release.ts
@@ -248,12 +248,11 @@ export default class ReleaseBinaryCommand extends AppCommand {
     return releaseRequestResponse.result;
   }
 
-  private async putReleaseDetails(client: AppCenterClient, app: DefaultApp, releaseId: number, releaseNotesString?: string, enabled?: boolean): Promise<models.ReleaseUpdateResponse> {
+  private async putReleaseDetails(client: AppCenterClient, app: DefaultApp, releaseId: number, releaseNotesString?: string): Promise<models.ReleaseUpdateResponse> {
     try {
       const { result, response } = await out.progress(`Updating release details...`,
         clientRequest<models.ReleaseUpdateResponse>(async (cb) => client.releases.updateDetails(releaseId, app.ownerName, app.appName, {
           releaseNotes: releaseNotesString,
-          enabled: enabled
         }, cb))
       );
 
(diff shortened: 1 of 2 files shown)


src/commands/distribute/release.ts
```diff
@@ -268,7 +269,46 @@ export default class ReleaseBinaryCommand extends AppCommand {
...
+        client.distributionGroups.get(this.app.ownerName, this.app.appName, destination, cb);
+      });
+
+      return result;
+    } catch (error) {
+      if (error.statusCode === 404) {
+        throw failure(ErrorCodes.InvalidParameter, `Could not find group ${destination}`);
+      } else {
+        debug(`Failed to distribute the release - ${inspect(error)}`);
+        throw failure(ErrorCodes.Exception, `Could not add group ${destination} to release ${releaseId}`);
```
owenniblock: I suggest changing the error to just express that there was a failure getting the distribution groups and remove the releaseId as an input as it's not needed here.

src/commands/distribute/release.ts
```diff
@@ -268,7 +269,46 @@ export default class ReleaseBinaryCommand extends AppCommand {
         throw failure(ErrorCodes.Exception, `failed to set distribution group and release notes for release ${releaseId}`);
       }
     }
+  }
+
+  private async getDistributionGroup(client: AppCenterClient, releaseId: number, destination: string): Promise<models.DistributionGroupResponse> {
```
owenniblock: Similar code is used in `add-destination.ts` - I suggest we refactor this in to a help file along with any other functions for accessing the distribution groups api. I guess this is made harder by the fork/PR stuff we're having to use right now - I'm guessing it might be easier to merge all this and then have a refactor PR that sort it out?

hagould: do we have any existing pattern for code sharing across commands?

owenniblock: Yes - if you make sure it's in a `lib` folder it shouldn't get picked up by the command / help creation routines.

Cross command I would put it in:
src/commands/distribute/lib

Our convention seems to be using either `*-helper.ts` or `*-util.ts`

owenniblock: You can also share amongst beacons by putting the file in /src/commands/lib - but I don't think you need to do that here.

PR #558, opened 2019-03-18
Stdover/show all destination

Stephanie-Dover:
View all release destinations via the CLI: Users can now view all destinations, including Testers, Distribution Groups, and Stores when viewing the release using the CLI

+ Added a message to display when there are no release destinations.
+ Showing all destinations for a release

### DIFF
--- a/src/commands/distribute/releases/show.ts
+++ b/src/commands/distribute/releases/show.ts
@@ -1,9 +1,8 @@
 import { AppCommand, CommandResult, ErrorCodes, failure, help, success, shortName, longName, required, hasArg } from "../../../util/commandline";
 import { AppCenterClient, models, clientRequest } from "../../../util/apis";
 import { out } from "../../../util/interaction";
-import { inspect, } from "util";
+import { inspect } from "util";
 import * as _ from "lodash";
-import { release } from "os";
 
 const debug = require("debug")("appcenter-cli:commands:distribute:releases:show");
 
--- a/src/commands/distribute/releases/show.ts
+++ b/src/commands/distribute/releases/show.ts
@@ -17,7 +17,6 @@ export default class ShowReleaseDetailsCommand extends AppCommand {
 
   public async run(client: AppCenterClient): Promise<CommandResult> {
     const app = this.app;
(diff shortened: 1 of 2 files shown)


test/commands/distribute/releases/show-test.ts
```diff
@@ -44,22 +44,45 @@ describe("releases show command", () => {
     Nock.enableNetConnect();
   });
 
-  describe("when everything works as expected", () => {
+  describe.only("when everything works as expected", () => {
```
pfleidi: We should remove this `.only()`, otherwise we'll disable all other tests in the test suite.

src/commands/distribute/releases/show.ts
```diff
@@ -61,7 +61,7 @@ export default class ShowReleaseDetailsCommand extends AppCommand {
       ["Download URL", "downloadUrl"],
       ["Install URL", "installUrl"],
       ["Icon URL", "appIconUrl"],
-      ["Distribution Group", "distributionGroups", (distributionGroups: models.DistributionGroup[]) => _.get(distributionGroups, "[0].name", "")]
+      ["Destinations", "destinations", (destinations: models.Destination[]) => destinations && destinations.length > 1 ? JSON.stringify(destinations, null, 2) : noDestinations]
```
pfleidi: Just out of curiosity: We are printing the destinations by using `JSON.stringify()` here. Is that a common pattern in the CLI? I would have expected a more reader-friendly format here.

DangMH: @pfleidi the current out.ts formatting options don't handle lists of objects particularly well.  Currently there is `out.report.allAttributes` which doesn't go further than one level, and `out.report.inspect` which doesn't do any delimiter formatting.

There also seems to be some report formatting that overrides any new line indentation.

At this point we considered adding a stringify wrapper to out.ts like inspect, but thought against it.

test/commands/distribute/releases/show-test.ts
owenniblock: Good catch!

PR #565, opened 2019-04-02
Fix getting symbol groups

pfleidi:
At the moment, when sending a request to get the symbol groups for an app, we are sending a `top` parameter of the value `Number.MAX_SAFE_INTEGER`.

Unfortunately this number is 9007199254740991 - which is way larger than the value accepted by the backend data store used by the symbol indexing service.

I’ve changed it to use a 32 bit integer to fix this issue.


ranterle:
What is the effect of this too large value being used? Does it cause an error or unexpected processing of the input?

pfleidi:
@ranterle The effect is that we get back an HTTP 400 error from Azure Table storage that tells us that the parameter that we provided to a query is not an Integer. ;)

PR #571, opened 2019-04-23
Add support for environment variable during app creation and app update.

bryansmi:


src/commands/apps/create.ts
```diff
@@ -46,6 +46,12 @@ export default class AppCreateCommand extends Command {
   @hasArg
   platform: string;
 
+  @help("The environment of the app. This must be one word, alphanumeric, first letter capitalized.")
```
owenniblock: If it's a requirement that the first letter should be capitalized - we should either check that here - or reformat the input and warn the user.

bryansmi: Pushed a commit which should address this by adding client-side validation.

owenniblock:
@bryansmi - we've been discussing this over here this morning and we're a bit worried about the wording (both of the help text for the parameter and the name of the parameter). I realise we're late in the day to this & we're holding up your User Story so very sorry to be a bottleneck!

I'm going to pass this over to @jmoody for discussion - we think it's important to try and get this right so it's not confusing for users.

src/commands/apps/create.ts
```diff
@@ -46,6 +47,12 @@ export default class AppCreateCommand extends Command {
...
```
jmoody: The spec says the acceptable values for this field are: Alpha, Beta, Production, Store, Enterprise.  Mentioning those here will help the user.
  
Instead of "The environment of the app", I think "The app release type" is better.

src/commands/apps/create.ts
```diff
@@ -46,6 +47,12 @@ export default class AppCreateCommand extends Command {
...
+  @shortName("e")
+  @longName("environment-app")
```
jmoody: This should be reversed: "app-environment" or "release-type"

src/commands/apps/lib/app-environment-validation.ts
```diff
@@ -0,0 +1,10 @@
+export const APP_ENVIRONMENT_VALIDATIONS = {
+  maxLength: {
+    rule: 100,
+    errorMessage: `environment-app length must be 100 characters or less`
+  },
+  matchRegexp: {
+    rule: /^$|^[A-Z0-9][a-z0-9]*$/,
+    errorMessage: `environment-app must be a single word starting with a capital letter or number, followed by lowercase`
+  }
+};
```
jmoody: Shouldn't we validate against the only accepted values: "Alpha, Beta, Production, Store, Enterprise"?

The error message should be explicit about what values are expected.

src/commands/apps/create.ts
bryansmi: > The spec says the acceptable values for this field are: Alpha, Beta, Production, Store, Enterprise. Mentioning those here will help the user.
Unfortunately, there are more than just these set of acceptable values. Those are the _suggested_ values - but a user can enter _any_ environment that meets the ruleset.

I'm not opposed to mentioning the values for this field - but they will make the error message significantly wordier. Is that an issue?

> Instead of "The environment of the app", I think "The app release type" is better.
I think this is problematic. HockeyApp has an existing field called `release_type` and we want to differentiate an App Center `environment` from that.

Perhaps 'The app environment...' is a better sentence starter?

Pre-emptively, I'd like to suggest something like:

'The app environment. Suggested values are Alpha, Beta, Production, Store, Enterprise. Custom values are allowed, and must be must be one word, alphanumeric, first letter capitalized.'

Thoughts?

cc: @blparr

src/commands/apps/create.ts
bryansmi: I think `app-environment` is probably acceptable. @blparr please object if you have an issue with that :)

bryansmi:
No worries about holding up the user story - let's get it right :) Added some responses for discussion with @jmoody

src/commands/apps/lib/app-environment-validation.ts
bryansmi: See above comment - custom values are permitted, those are just suggestions :D

blparr:
Looks good.

bryansmi:
Pushed some changes that hopefully will resolve your concerns. I'll try and check back first thing tomorrow morning and sync to resolve any further issues.

src/commands/apps/create.ts
jmoody: The spec says that no custom values are allowed.

bryansmi:
Closing this PR - will be creating a new one to support the change from `environment` to `release_type`.

PR #572, opened 2019-04-24
Add support for NUnit3 XML reports from UITest

owenniblock:
### Motivation

`NUnit3` test runs against the latest `UITest` fail when they attempt to merge the output XML (this is because of a bug in the merging on the pipeline but also because of the code here which needed updating to match).

This PR adds support for the new `NUnit3` report format.

### Caveat

Looking at the existing code, I suspect that there are some issues merging result states between test runs (both in `NUnit2` and `NUnit3`) - I have not had time to investigate this thoroughly but I believe it would manifest if running on several devices and only a subset of the devices fail.

Example CLI output:

```
Current test status: Running on 1 device (3 / 4 completed, 0 pending) - Report 50% done

Current test status: Tests completed. Processing data.

Current test status: Tests completed. Processing data.

Current test status: Tests completed. Processing data.

Current test status: Done!
Total scenarios: 4
4 passed
0 failed
Total steps: 6


Downloaded artifacts to /Users/owenniblock/tmp/nunit_xml_zip.zip
```

@maxxkrakoa - I'm not sure who's best to review this PR. Normally I'd ask Simon - let me know!


src/commands/test/lib/nunit-xml-util.ts
```diff
@@ -27,7 +27,11 @@ export class NUnitXmlUtil extends XmlUtil {
         self.removeEmptySuitesTransformation(xml);
 
         if (mainXml) {
-          mainXml = self.combine(mainXml, xml);
+          if (this.isNUnit3(xml)) {
+          mainXml = self.combineNUnit3(mainXml, xml);
```
maxxkrakoa: Should be indented

src/commands/test/lib/nunit-xml-util.ts
```diff
@@ -106,28 +129,51 @@ export class NUnitXmlUtil extends XmlUtil {
   }
 
   combineTestResultsAttribute(xml1: Document, xml2: Document, attributeName: string) {
```
maxxkrakoa: Should this method be renamed to `combineResultsAttribute` to fit in with the other renamed methods?

src/commands/test/lib/nunit-xml-util.ts
```diff
@@ -106,28 +129,51 @@ export class NUnitXmlUtil extends XmlUtil {
...
+    let testResults: Element[] = this.collectAllElements(xml.documentElement, "test-results");
+    if (testResults.length === 0) {
+      testResults = this.collectAllElements(xml.documentElement, "test-run");
+    }
+
+    return testResults[0];
+  }
+
+  isNUnit3(xml: Document): boolean {
+    const testResults: Element[] = this.collectAllElements(xml.documentElement, "test-results");
```
john7doe: this is more of a not nunit2 test. Perhaps its better to test for precense of "test-run" instead of absence of "test-results"

src/commands/test/lib/nunit-xml-util.ts
```diff
@@ -106,28 +129,51 @@ export class NUnitXmlUtil extends XmlUtil {
...
+    if (testResults.length === 0) {
```
john7doe: remove if and return false can be simplified to

return testResults.length === 0

src/commands/test/lib/nunit-xml-util.ts
```diff
@@ -62,6 +66,25 @@ export class NUnitXmlUtil extends XmlUtil {
...
+    this.combineTestResultsAttribute(xml1, xml2, "passed");
+    this.combineTestResultsAttribute(xml1, xml2, "failed");
+    this.combineTestResultsAttribute(xml1, xml2, "inconclusive");
+    this.combineTestResultsAttribute(xml1, xml2, "skipped");
+    this.combineTestResultsAttribute(xml1, xml2, "asserts");
+
+    const testSuitesParent: Element = this.collectAllElements(xml1.documentElement, "test-run")[0];
+    const testSuites: Element[] = this.collectChildren(xml2.documentElement, "test-suite");
+
+    testSuites.forEach((child: Element) => {
```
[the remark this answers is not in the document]
owenniblock: Nice!

src/commands/test/lib/nunit-xml-util.ts
owenniblock: I've done it like this pending NUnit4 coming out so that it's easier to convert to:
> getNUnitVersion: int
Or something similar. Although it might not be the case, I figured NUnit4 is more likely to follow similar report format the NUnit3 - at which point we may need to dig in to the version attribute on the `test-results` node.

src/commands/test/lib/nunit-xml-util.ts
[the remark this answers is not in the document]
owenniblock: I couldn't get this to work:

`.appendChild` isn't compatible with `...` because it expects at least 1 child element [this error gets raised](https://github.com/Microsoft/TypeScript/issues/28010)

`.append` appears to not work for an `Element` which is confusing. Unless you're super keen for me to do this I suggest we don't invest _too_ much time in it.

owenniblock:
@john7doe - I've not done 2 of your suggestions - let's discuss tomorrow if you think they're important enough to do!

owenniblock:
Merging despite CI build error - will handle that separately.

PR #574, opened 2019-04-29
Add source-version argument to the build queue command

CodyBrouwers:
Whenever my company increments the version of our app, we trigger a build on AppCenter which works great. The issue this PR aims to solve is that once the build is complete, the notification we get in Slack does not include the version notes that are in the commit message. I've asked about getting this fixed on the AppCenter Intercom for a while now but was told that its unfortunately not a priority. Jihye did mention there is a temporary fix possible by manually making the API POST to AppCenter's API and including the `sourceVersion` argument  as seen here: https://openapi.appcenter.ms/#/build/builds_create

This works great so this PR adds the `sourceVersion` argument to the `build queue` command so it is possible as a workaround for now until AppCenter is able to pick up the commit message that triggered the build automatically.

Thanks and let me know if there are any required changes to this PR!

Cody Brouwers

### DIFF
--- a/src/commands/build/queue.ts
+++ b/src/commands/build/queue.ts
@@ -9,7 +9,7 @@ const debug = require("debug")("appcenter-cli:commands:build:queue");
 @help("Queue a new build")
 export default class QueueBuildCommand extends AppCommand {
 
-  @help("Branch to be build")
+  @help("Branch to be built")
   @shortName("b")
   @longName("branch")
   @required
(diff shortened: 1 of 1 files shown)


owenniblock:
We should be releasing latest cli in a few days (all being well) so this will be rolled in to that release.

CodyBrouwers:
Sound good @owenniblock , thanks! 😁

PR #576, opened 2019-04-30
App release type option

bryansmi:
Add the ability to set a release type when creating or updating an app.


bryansmi:
Updated generated models from current swagger. `environment` is now completely deprecated.

PR #580, opened 2019-05-03
Support UITest 3 & NUnit 3

jonstoneman:
- Update exe name when using locally built test-cloud.exe
 - Use UITest 3.0.0 and NUnit 3.11.0


src/commands/test/lib/uitest-preparer.ts
```diff
@@ -193,7 +193,7 @@ export class UITestPreparer {
     let testCloudPath = path.join(toolsDir, "test-cloud.exe");
 
     if (!await fileExists(testCloudPath)) {
-      testCloudPath = path.join(toolsDir, "Xamarin.UITest.Console.exe");
+      testCloudPath = path.join(toolsDir, "Xamarin.UITest.CLI.exe");
```
owenniblock: Good catch!

PR #593, opened 2019-06-04
Set Current is only compatible with `login` - tell users in help text

owenniblock:
Improves the error message when user tries to set the app and isn't using login.

I'm also creating an internal bug to change the environment variable `MOBILE_CENTER_CURRENT_APP` to `APP_CENTER_CURRENT_APP`.


owenniblock:
@jmoody - I think you said you would approve this after our chat :) Wanted to double check when you've got a min.

PR #594, opened 2019-06-07
Add command for uploading Android mappings

ranterle:
Adds the `crashes upload-mappings` command with arguments
```
--mapping|-m -> for the mapping file name
--version-name|-n -> for the version name (marketing version i.e. "1.0")
--version-code|-c -> for the version code (build number, i.e. "1")
```

1. All three arguments are required. The version name and version code are required in App Center Diagnostics to associate the mappings with the right published app version
1. Android ProGuard mappings are directly uploaded as a mapping.txt file
1. Internally changed the upload of symbols to use the generated `UploadSymbolsBeginRequest` model and generified the upload symbols helper functionality to remove references to only uploading zip files as symbol artifacts


src/commands/crashes/upload-mappings.ts
```diff
@@ -0,0 +1,88 @@
...
+  }
+
+  private getLowerCasedFileExtension(filePath: string): string {
+    return Path.extname(filePath).toLowerCase();
+  }
+
+  private validateParameters() {
+    // check that user have selected all of --mapping, --version-name, and --version-code
+    if (_.isNil(this.mappingPath) || _.isNil(this.versionName) || _.isNil(this.versionCode)) {
+      throw failure(ErrorCodes.InvalidParameter, "all of '--mapping', '--version-name', and '--version-code' are required");
```
bryansmi: Just a thought that I haven't had a chance to dive further into: do we typically group error messages together in the CLI? Or do we fail on each specific parameter that is missing?

Usually, when I use a CLI tool, I'm told about each specific parameter that is missing/invalid. However, I'm not sure that's a better UX than this.

test/commands/crashes/upload-mappings/upload-mappings-test.ts
```diff
@@ -0,0 +1,244 @@
...
+  });
+
+  beforeEach(() => {
+    tmpFolderPath = Temp.mkdirSync("uploadMappingsTest");
+    postSymbolSpy = Sinon.spy();
+    patchSymbolSpy = Sinon.spy();
+    abortSymbolUploadSpy = Sinon.spy();
+  });
+
+  describe("input validation", () => {
```
bryansmi: Some additional test ideas:
- If we only want to support `mapping.txt` as a valid filename, we could test for that here
- If `version` (I believe) _must_ be a number, we should check that we can cast the string input to a number

test/commands/crashes/upload-mappings/upload-mappings-test.ts
```diff
@@ -0,0 +1,244 @@
...
+    return await uploadSymbolsCommand.execute();
+  }
+
+  function testCommandSuccess(result: CommandResult, executionScope: Nock.Scope, abortScope: Nock.Scope) {
+      expect(result.succeeded).to.eql(true, "Command should be successfully completed");
+      expect(abortScope.isDone()).to.eql(false, "Upload should not be aborted");
+      executionScope.done(); // All normal API calls are executed
+  }
+
+  function testUploadFailure(result: CommandResult, executionScope: Nock.Scope, patchScope: Nock.Scope) {
```
bryansmi: If this isn't being re-used, perhaps we should move it back into the test for the time being?

src/commands/crashes/upload-mappings.ts
```diff
@@ -0,0 +1,88 @@
...
+
+@help("Upload the Android mappings for the application")
+export default class UploadMappings extends AppCommand {
+  @help("Path to an Android mapping.txt file.")
+  @shortName("m")
+  @longName("mapping")
+  @hasArg
+  public mappingPath: string;
+
+  @help("The version name to associate with the mappings.")
```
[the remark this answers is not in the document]
ranterle: The name in an Android project for this is actually "version name" or "versionName" in the manifest so I think the language works here.

src/commands/crashes/upload-mappings.ts
```diff
@@ -0,0 +1,88 @@
...
+  @shortName("n")
+  @longName("version-name")
+  @hasArg
+  public versionName: string;
+
+  @help("The version code to associate with the mappings.")
```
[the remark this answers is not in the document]
ranterle: The name in an Android project for this is actually "version code" or "versionCode" in the manifest so I think the language works here.

src/commands/crashes/upload-mappings.ts
```diff
@@ -0,0 +1,88 @@
...
+  @hasArg
+  public versionCode: string;
+
+  public async run(client: AppCenterClient): Promise<CommandResult> {
+    const app: DefaultApp = this.app;
+
+    this.validateParameters();
+
+    this.getStatsForFsPath(this.mappingPath);
+    if (this.getLowerCasedFileExtension(this.mappingPath) !== ".txt") {
```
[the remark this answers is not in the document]
ranterle: I think checking for the entire file name might be a bit restrictive, especially since the CLI is also executed on a developer machine, not only in the concept of one particular build which creates one mapping.txt file - so the user might have assembled a different structure, i.e. a folder where they keep multiple mapping.txt files for different versions like "mapping-1.0-21.txt" etc.

src/commands/crashes/upload-mappings.ts
```diff
@@ -0,0 +1,88 @@
...
+
+  private getStatsForFsPath(filePath: string): Fs.Stats | null {
+    // take fs entry stats (and check it's existence BTW)
+    try {
+      debug(`Getting FS statistics for ${filePath}`);
+      return Fs.statSync(filePath);
+    } catch (error) {
+      if (error.code === "ENOENT") {
+        // path points to non-existing file system entry
+        throw failure(ErrorCodes.InvalidParameter, `path ${filePath} points to non-existent item`);
```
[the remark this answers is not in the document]
ranterle: Made the wording here consistent with the rest of actions, i.e. when looking at the `upload_symbols` command, this has similar wording - I'm no big fan either but didn't want to sacrifice consistency here.

src/commands/crashes/upload-mappings.ts
```diff
@@ -0,0 +1,88 @@
...
+      } else {
+        // other errors
+        debug(`Failed to get statistics for file system entry ${filePath} - ${inspect(error)}`);
+        throw failure(ErrorCodes.Exception, `failed to get statistics for file system entry ${filePath}`);
```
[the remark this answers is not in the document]
ranterle: Same as above, like your preference, but kept it consistent with other log and error statements in the codebase.

src/commands/crashes/upload-mappings.ts
```diff
@@ -0,0 +1,88 @@
...
+      }
+    }
+  }
+
+  private getLowerCasedFileExtension(filePath: string): string {
```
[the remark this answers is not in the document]
ranterle: This is used for finding the filename extension in line 44.

src/commands/crashes/upload-mappings.ts
ranterle: I agree, that it might be superfluous to notify the user if they did not provide one out of three parameters about the other two which they did indeed provide. I also think the error message can be improved as it only references the `--` parameters but not their `-` variants.

test/commands/crashes/upload-mappings/upload-mappings-test.ts
ranterle: Yes on the account that `version code` needs to be a positive (non-zero) integer, will add a validation here. Typically this should get caught by the upload pipeline, but it is a restriction imposed by the Android SDK, not App Center, so we can require it in here, too.

src/commands/crashes/upload-mappings.ts
ranterle: On the contrary I think this kind of message delivers the most concise requirement alas "you need to provide x y and z" - added the short names of the switches - what do you think?

test/commands/crashes/upload-mappings/upload-mappings-test.ts
ranterle: Sounds good to me.

src/commands/crashes/upload-mappings.ts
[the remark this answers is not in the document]
bryansmi: Do we need a 1-line helper function that isn't reused elsewhere? I think we could just directly call `Path.extname(this.mappingPath).toLowerCase()` in Line 44, no?

src/commands/crashes/upload-mappings.ts
bryansmi: I like this.

src/commands/crashes/upload-mappings.ts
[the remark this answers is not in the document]
ranterle: Fair enough - I didn't read your message as "are we reusing this" but just "are we using this" - I don't have strong feelings here, it is another kind where we took the same approach as `upload-symbols` but since it's private, we can just inline.

src/commands/crashes/upload-mappings.ts
```diff
@@ -0,0 +1,86 @@
...
+  public versionCode: string;
+
+  public async run(client: AppCenterClient): Promise<CommandResult> {
+    const app: DefaultApp = this.app;
+
+    this.validateParameters();
+
+    this.getStatsForFsPath(this.mappingPath);
+    if (Path.extname(this.mappingPath).toLowerCase() !== ".txt") {
+      throw failure(ErrorCodes.InvalidParameter, `path ${this.mappingPath} does not point to valid mapping.txt file`);
```
DennisPan: Can we be more explicit in this message? It'd be helpful to point out that we require file to have `.txt` as extension

ranterle: Yeah, I can see how it might be improvable.

README.md
```diff
@@ -72,6 +72,7 @@ Below is the list of commands currently supported by Visual Studio App Center CL
 | | |
 | `appcenter crashes upload-missing-symbols` | Upload missing crash symbols for the application (only from macOS) |
 | `appcenter crashes upload-symbols` | Upload the crash symbols for the application |
+| `appcenter crashes upload-mappings` | Upload Android mapping files for the application |
```
owenniblock: This file is semi-automatically updated as part of the release process - you don't need to change it on the feature PRs :)

PR #625, opened 2019-07-30
Update UITest in launch tests

jonstoneman:


owenniblock:
It looks like you also need to update a test for the build to succeed.

PR #626, opened 2019-08-01
Added build_version optional parameter to CLI release command

snowpardx:
**Why:** we need to support distribution of WinForms/WPF apps from .msi and .zip files
**What/How:** 
1. added `--build-version` extra parameter to `distribute release` command 
2. updated swagger (only build_version change and not the whole description from bifrost)
3. generated autorest models
4. passed the buildVersion to `release_uploads` endpoint call

### DIFF
--- a/src/util/apis/generated/operations/index.d.ts
+++ b/src/util/apis/generated/operations/index.d.ts
@@ -13828,7 +13828,7 @@ export interface ReleaseUploads {
      * @param {number} [options.releaseId] The ID of the release.
      *
      * @param {string} [options.buildVersion] The build version of the uploaded
-     * WPF/Winforms
+     * binary
      *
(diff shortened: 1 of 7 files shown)


src/util/apis/generated/operations/index.d.ts
```diff
@@ -13849,6 +13852,9 @@ export interface ReleaseUploads {
      *
      * @param {number} [options.releaseId] The ID of the release.
      *
+     * @param {string} [options.buildVersion] The build version of the uploaded
+     * WPF/Winforms
```
cainejette: to align with `releases` swagger, let's say `The build version of the uploaded binary`

snowpardx: Thanks for heads up, missed that we updated the text. Updated swagger file used and regenerated models

PR #631, opened 2019-08-07
Update CodePush release flow

cjonsmith:
As a part of the general CodePush overhaul, we've updated the process to publish a new release. For now, both flows are supported, but we'll be updating our clients to only target the new release flow.

This PR does not affect the usage of the `codepush release*` commands.


owenniblock:
I've concerned about why we needed to roll back the swagger update - this would suggest that things are broken and need to be fixed (if we get in to the habit of manually editing the auto-generated files that will cause problems moving forwards).

My gut reaction is to merge this, and then investigate the swagger issue before we release.

Was the swagger update performed production? Or another environment?

cjonsmith:
This was running against production, and yeah, I agree that we should follow up with a fix to address the autorest generation issues that we ran into during our work for this change.

We won't let this block our PR though as you suggested. We can follow up later with a change that updates the effected parts of the CLI after a swagger client generation.

PR #637, opened 2019-08-20
Feature/uwp support

Vahagn1993:
Changes:
1. Added `.appxsym` file support for UWP Apps
2. Updated mutually exclusion logic in `upload_symbols` command

### DIFF
--- a/src/commands/crashes/upload-symbols.ts
+++ b/src/commands/crashes/upload-symbols.ts
@@ -183,13 +183,12 @@ export default class UploadSymbols extends AppCommand {
       return `${commaJoinedArgs} ${cnj} ${_.last(args)}`;
     };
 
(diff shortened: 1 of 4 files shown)


src/commands/crashes/upload-symbols.ts
```diff
@@ -54,6 +54,10 @@ export default class UploadSymbols extends AppCommand {
   @hasArg
   public breakpadPath: string;
 
+  @help("Path to an appxsym file containing UWP symbols.")
+  @longName("appxsym")
+  @hasArg
+  public appxSymPath: string;
```
owenniblock: Please add a newline after this line.

Vahagn1993: Done!

Issue #592, opened 2019-05-30
Releasing directly to Android store Production suddenly doesn't work anymore


devedse:
Hi all,

In december last year I fully automated the release of my Android application to the App Store using AppCenter. See for example this build:
https://ci.appveyor.com/project/devedse/deveblockstacker/builds/21288368

However when I just triggered a new build it failed with the error:
`Error: Could not find group Production`

I'm now trying to investigate what is going on, I tried executing the following command locally:
```
appcenter distribute release -g Production -f "C:\XKanWeg\Block\deveblockstacker_android.deveblockstacker_android-Signed (1).apk" -a Devedse/DeveBlockStacker.Android -r "This is a test" --token ****************
```

However this now also returns the error:
`Error: Could not find group Production`

![image](https://user-images.githubusercontent.com/2350015/58654034-53819b00-8317-11e9-9b33-d4f41470025c.png)

Does anyone have an idea on how to fix this?

```
C:\XKanWeg\Block>appcenter distribute release -g Production -f "C:\XKanWeg\Block\deveblockstacker_android.deveblockstacker_android-Signed (1).apk" -a Devedse/DeveBlockStacker.Android -r "This is a test" --token ******************************** --debug
Using appcenter-cli version: 1.1.18
| Loading prerequisites...logFilter, request: {
  "rawResponse": false,
  "queryString": {},
  "url": "https://api.appcenter.ms/v0.1/apps/Devedse/DeveBlockStacker.Android/distribution_groups/Production/members",
  "method": "GET",
  "headers": {
    "Content-Type": "application/json; charset=utf-8",
    "user-agent": "appcenterCli/1.1.18 NodeJS/v12.3.1 win32/10.0.17134",
    "x-api-token": "********************************"
  },
  "body": null
}
/ Loading prerequisites...Response status code: 404
Body: {"error":{"code":"NotFound","message":"Could not find distribution group Production in this app"}}
| Creating release upload...logFilter, request: {
  "rawResponse": false,
  "queryString": {},
  "url": "https://api.appcenter.ms/v0.1/apps/Devedse/DeveBlockStacker.Android/release_uploads",
  "method": "POST",
  "headers": {
    "Content-Type": "application/json; charset=utf-8",
    "user-agent": "appcenterCli/1.1.18 NodeJS/v12.3.1 win32/10.0.17134",
    "x-api-token": "********************************"
  },
  "body": null
}
| Creating release upload...Response status code: 201
[...]

PR #647, opened 2019-09-19
Add support to publish to stores

jpeg-ms:
Changes:
- Add command `distribute stores list`
- Add option `-s/--store` to the command `distribute release`


src/commands/distribute/stores/category.txt
```diff
@@ -0,0 +1 @@
+Commands to distribute to stores (App Store, Google Play, Intune)
```
jerietve: Remove "Commands to" to follow convention. Consider adding an object, the thing that you can distribute.

src/commands/distribute/stores/list.ts
```diff
@@ -0,0 +1,48 @@
+import { AppCommand, CommandResult, ErrorCodes, failure, help, success } from "../../../util/commandline";
+import { AppCenterClient, models, clientRequest, ClientResponse } from "../../../util/apis";
+import { out } from "../../../util/interaction";
+import { inspect } from "util";
+import * as _ from "lodash";
+
+const debug = require("debug")("appcenter-cli:commands:distribute:stores:list");
+
+@help("Lists all stores of the app")
```
jerietve: The help for apps list is "Get list of configured applications"
Let's make this one something like "Get list of stores configured for an app"

src/commands/distribute/stores/category.txt
jerietve: I see releases does this 'commands for' thing. It doesn't add anything though, just takes up space.

jerietve: nvm let's keep what you wrote until we do another doc cleanup

src/commands/distribute/stores/list.ts
```diff
@@ -0,0 +1,48 @@
...
+      debug(`Failed to get list of the stores - ${inspect(error)}`);
+      return failure(ErrorCodes.Exception, "failed to fetch list of all stores");
+    }
+
+    if (storesListRequestResponse.response.statusCode >= 400) {
+      return failure(ErrorCodes.Exception, "failed to fetch list of all stores");
+    }
+
+    const storesNames = _(storesListRequestResponse.result)
+      .sortBy((store) => [store.type, store.track, store.name])
```
jerietve: Consider doing the above lines once to avoid code duplication and multiple enumeration

src/commands/distribute/stores/list.ts
```diff
@@ -0,0 +1,48 @@
...
+      .map((store) => store.type).value();
+
+    const storesTracks = _(storesListRequestResponse.result)
+      .sortBy((store) => [store.type, store.track, store.name])
+      .map((store) => store.track).value();
+
+    const outputArray = _.zip(storesNames, storesTypes, storesTracks);
+
+    // Printing the result table
+    out.table(out.getCommandOutputTableOptions(["Store", "Type", "Track"]), outputArray);
```
jerietve: Consider showing a message saying there are no stores configured or something rather than an empty table when there are no stores.

src/commands/distribute/release.ts
```diff
@@ -11,7 +11,7 @@ import { getDistributionGroup, addGroupToRelease } from "./lib/distribute-util";
 
 const debug = require("debug")("appcenter-cli:commands:distribute:release");
 
-@help("Upload release binary and trigger distribution")
+@help("Upload release binary and trigger distribution, at least one of --store or --group must be specified")
```
jerietve: This CLI really needs the concept of PowerShell parameter sets..

src/commands/distribute/release.ts
```diff
@@ -81,17 +86,31 @@ export default class ReleaseBinaryCommand extends AppCommand {
...
     const releaseId = this.extractReleaseId(releaseUrl);
 
-    debug("Distributing the release");
-    await this.distributeRelease(client, app, releaseId, releaseNotesString);
+    debug("Setting release notes");
+    await this.putReleaseDetails(client, app, releaseId, releaseNotesString);
+
+    if (!_.isNil(this.distributionGroup)) {
+      debug("Distributing the release to a group");
+      await this.distributeRelease(client, app, releaseId);
```
jerietve: We can now have a situation where publishing to a distribution group succeeded, followed by a failure to publish to the store, or the other way around cause things are async, in which case you would only see the failure and not get confirmation of the success.

src/commands/distribute/stores/category.txt
jerietve: I see this is a unique situation in the CLI where a "command folder" (distribute) has both commands (release) and more command folders (groups, releases, stores) inside it. So the sentence makes sense.

jpeg-ms:
This should solve #592.

test/commands/distribute/stores/list-test.ts
```diff
@@ -0,0 +1,109 @@
...
+import * as _ from "lodash";
+
+use(ChaiAsPromised);
+
+import ListStoresCommand from "../../../../src/commands/distribute/stores/list";
+import { CommandArgs } from "../../../../src/util/commandline";
+import { out } from "../../../../src/util/interaction";
+import { ExternalStoreResponse } from "../../../../src/util/apis/generated/models";
+
+describe("releases show command", () => {
```
jerietve: stores list

src/commands/distribute/release.ts
jpeg-ms: The only thing that can happen is successful distribution to a group but failure at the store release.

src/commands/distribute/stores/category.txt
jpeg-ms: I agree but we can deal with these inconsistencies in another PR.

PR #649, opened 2019-09-19
Update Appium to latest

jpeg-ms:
This should solve the compliance issue


jpeg-ms:
Abandoning this. We can't update JUnit as it hasn't been tested on App Center Test yet, even though it is the component that triggers the "Component Governance" issue.

PR #652, opened 2019-09-26
Fix Hermes support for React Native 0.61.0

lpikora:
Fix Hermes support for React Native 0.61.0: change Hermes binary path from hermesvm to hermes-engine because hermesvm is now deprecated…

React Native 0.61.x start using hermes-engine instead of hermesvm. Introduced in: https://github.com/facebook/react-native/pull/25908/commits/70f1c87db120ad63fee7f86a8a10b3fde2df47d2

https://www.npmjs.com/package/hermesvm is now deprecated in favour of hermes-engine.


jpeg-ms:
Hi @lpikora 
Thanks a lot for your contribution, but we have another PR that was opened later to solve the same issue: #654. That one supports both engines so we'll prefer it for now.

PR #654, opened 2019-09-26
Prefer hermes-engine over hermesvm

### LINKED ISSUE #653: CodePush releases with hermes do not respect `hermesCommand`
The current appcenter-cli  assumes that the hermes binary lives under `node_modules/hermesvm` which is not true for react native >= 0.61.x, it is now at `node_modules/hermes_engine`.

Even more problematic, if a react native < 0.61.x has configured their `hermesCommand` to use `hermes_engine` instead of `hermesvm` (to use the newer hermes version), the appcenter cli will use the old `hermesvm` binary, generating incompatible bytecode that causes an immediate crash and rollback when published.

The cli should use the configured `hermesCommand` when generating a codepush release.

azundo:
If hermes-engine is installed in node_modules, prefer the hermes binary
from there vs the one in hermesvm.

Ideally the `hermesCommand` could be parsed and used from the build.gradle
but parsing the gradle-to-js output for `hermesCommand` is not trivial. I think the
heuristic of preferring `hermes-engine` over `hermesvm` should be correct in the
vast majority of cases.

I've tested this locally with `ts-node` on a react-native 0.60.x project for the case when both `hermes-engine` and `hermesvm` are installed and when only `hermesvm` is installed.

Closes #653

### DIFF
--- a/src/commands/codepush/lib/react-native-utils.ts
+++ b/src/commands/codepush/lib/react-native-utils.ts
@@ -271,8 +271,9 @@ export function runHermesEmitBinaryCommand(bundleName: string, outputFolder: str
   }
 
   out.text(chalk.cyan("Converting JS bundle to byte code via Hermes, running command:\n"));
-  const hermesProcess = spawn(path.join("node_modules", "hermesvm", getHermesOSBin(), "hermes"), hermesArgs);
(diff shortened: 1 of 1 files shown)


jpeg-ms:
Thanks for your contribution!

PR #655, opened 2019-09-27
Add dedicated commands for publishing to stores and groups

jerietve:
Add two new commands:

- `distribute stores publish` which performs the same actions as `distribute release --store`.
- `distribute groups publish` which performs the same actions as `distribute release --group`.


jpeg-ms:
Dropping this as it has become a mess to review. A new PR will be opened with the same actual changes.

PR #659, opened 2019-10-01
Fix "npm pack" usage

jpeg-ms:
`npm pack` can be used to generate a smaller package, that we can then publish on GitHub. The rules in this PR select only the files in `bin/` and `dist/` for installation.


jerietve:
If they don't select the files in the root, then the rule in .npmignore may not be necessary.

jpeg-ms:
> If they don't select the files in the root, then the rule in .npmignore may not be necessary.
That's right, I added it for consistency with .gitignore.

PR #660, opened 2019-10-01
Update azure-pipelines.yml to ignore false alerts

jpeg-ms:
Add optional variable to ignore some directories for Component Governance. Set up to ignore templates using a variable set in AzDO.

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -17,7 +17,7 @@ jobs:
   - script: |
       npm ci
       npm run build
-    displayName: 'npm install and build'
+    displayName: 'Install and build'
 
   - task: ComponentGovernanceComponentDetection@0
     displayName: Component Governance (PR)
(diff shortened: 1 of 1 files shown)


azure-pipelines.yml
```diff
@@ -48,26 +49,12 @@ jobs:
   - task: CopyFiles@2
     inputs:
       sourceFolder: '$(Build.SourcesDirectory)'
-      contents: '*.tgz' 
+      contents: '*.tgz'
       targetFolder: $(Build.ArtifactStagingDirectory)/npm
     displayName: 'Copy npm package'
 
-  - task: CopyFiles@2
```
jerietve: No need for these files? Do you know why this was here? Unrelated to the other PR, just cleanup?

jpeg-ms: The folder structure was slightly different. Basically package-lock was manually copied over, but instead I made sure to add it to the package created by `npm pack`, so we take it from there. So yeah, it's related to the other PR (already merged).

PR #665, opened 2019-10-02
Add base for functional tests

jerietve:
Functional tests are in /test/functional. The Main.ps1 PowerShell script installs Pester and runs all files calls *Tests.ps1 in that directory and any subfolders of that directory. It runs against the build in /dist by overriding the `appcenter` command by temporarily appending the /test/functional/bin folder to `PATH`.

Results are stored in testresult.xml and analyzed by an Azure Pipelines task to prevent release with failures.


jpeg-ms:
This is the beginning of a series of integration tests that should be able to call real `appcenter-cli` commands. This uses "Pester" as a test suite, in Powershell.
Eventually we would use private credentials (a secret token) setup inside our Azure DevOps pipeline. The test results are accessible by the community within GitHub even though the full output on Azure DevOps is unfortunately not visible.

PR #674, opened 2019-10-09
modified release.ts to utilize Stream instead of Buffer for uploading a file.

KojiNakamaru:
A large file over 2GB cannot be uploaded as the current implementation reads the whole content into memory as Buffer. This modification utilizes Stream instead of Buffer for uploading a file.

cf. https://github.com/capaj/how-to-node-io

### DIFF
--- a/src/commands/distribute/release.ts
+++ b/src/commands/distribute/release.ts
@@ -8,6 +8,8 @@ import * as Path from "path";
 import * as Pfs from "../../util/misc/promisfied-fs";
 import { DefaultApp } from "../../util/profile";
 import { getDistributionGroup, addGroupToRelease } from "./lib/distribute-util";
+import * as fs from "fs";
+import * as stream from "stream";
 
 const debug = require("debug")("appcenter-cli:commands:distribute:release");
 
@@ -60,7 +62,7 @@ export default class ReleaseBinaryCommand extends AppCommand {
     this.validateParameters();
 
     debug("Loading prerequisites");
-    const [distributionGroupUsersCount, storeInformation, releaseBinaryFileBuffer, releaseNotesString] = await out.progress("Loading prerequisites...", this.getPrerequisites(client));
+    const [distributionGroupUsersCount, storeInformation, releaseBinaryFileStream, releaseBinaryFileStats, releaseNotesString] = await out.progress("Loading prerequisites...", this.getPrerequisites(client));
 
     this.validateParametersWithPrerequisites(storeInformation);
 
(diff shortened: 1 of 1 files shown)


jpeg-ms:
Thanks for your contribution! We'll have a closer look soon.

For now, it seems the unit tests have been broken. (I'll also have to check why the results aren't exposed here on GitHub).
```
  74 passing (533ms)
  3 failing

  1) release command
       build-version
         when validates input arguments
           passes with aab for stores:
     Uncaught Error: ENOENT: no such file or directory, open 'valid.aab'
  

  2) release command
       build-version
         when validates input arguments
           passes with ipa for stores:
     Uncaught Error: ENOENT: no such file or directory, open 'valid.ipa'
  

  3) release command
       build-version
         when validates input arguments
           passes with ipa for stores:
     Uncaught Error: ENOENT: no such file or directory, open 'valid.ipa'
```

KojiNakamaru:
I've committed 129b1de to pass those unit tests, could you please check it.

src/commands/distribute/release.ts
```diff
@@ -181,12 +184,25 @@ export default class ReleaseBinaryCommand extends AppCommand {
...
 
-    return Promise.all([distributionGroupUsersNumber, storeInformation, fileBuffer, releaseNotesString]);
+    return Promise.all([distributionGroupUsersNumber, storeInformation, fileStream, fileStats, releaseNotesString]);
   }
 
-  private async getReleaseFileBuffer(): Promise<Buffer> {
+  private async getReleaseFileStream(): Promise<stream.Stream> {
     try {
-      return await Pfs.readFile(this.filePath);
+      fs.statSync(this.filePath);
```
jpeg-ms: Hmm I don't think this should be there, as we're ignoring the return value, and doing the stat twice.
How about having this function return both the stream and the fsstat?

src/commands/distribute/release.ts
```diff
@@ -162,9 +164,10 @@ export default class ReleaseBinaryCommand extends AppCommand {
     }
   }
 
-  private getPrerequisites(client: AppCenterClient): Promise<[number | null, models.ExternalStoreResponse | null, Buffer, string]> {
+  private getPrerequisites(client: AppCenterClient): Promise<[number | null, models.ExternalStoreResponse | null, stream.Stream, fs.Stats, string]> {
     // load release binary file
-    const fileBuffer = this.getReleaseFileBuffer();
+    const fileStream = this.getReleaseFileStream();
+    const fileStats = this.getReleaseFileStats();
```
jpeg-ms: (see below comment)
Here you could have:
```
const { fileStats, fileStream } = this.getReleaseFileStream();
```

KojiNakamaru: How about 1459af0 ?

jpeg-ms: Hey that's pretty cool, but isn't there an async/await version of those calls? Rather than using the sync call, since eventually we are doing a promise.all() we should be considering promises all the way :)

jerietve: Since this just creates a stream and doesn't actually read the file, async may not be necessary. It doesn't appear to exist, probably for that reason. It's still a file system interaction, just a relatively quick one.
I'm looking at newer code than where this comment was originally placed.

jpeg-ms: Apparently you're right, and I was confused by older code which did await.

KojiNakamaru:
As you noticed, I've updated the code further: 152688d . Could you please let me know if any change is required.

jpeg-ms:
> As you noticed, I've updated the code further: [152688d](https://github.com/microsoft/appcenter-cli/commit/152688d230eda6eb2ff2a559698f5c1aea58164a) . Could you please let me know if any change is required.

Yes. I'm on it. Hang on tight! :)

PR #675, opened 2019-10-10
release: Skip empty release notes

jpeg-ms:
This is allowed for Google Play and distribution groups.
This check avoids getting a 400 error which was then badly
translated as:
  "changing distribution group is not supported"

### DIFF
--- a/src/commands/distribute/release.ts
+++ b/src/commands/distribute/release.ts
@@ -91,8 +91,12 @@ export default class ReleaseBinaryCommand extends AppCommand {
     debug("Extracting release ID from the release URL");
     const releaseId = this.extractReleaseId(releaseUrl);
 
-    debug("Setting release notes");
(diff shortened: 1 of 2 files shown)


src/commands/distribute/release.ts
```diff
@@ -349,6 +353,7 @@ export default class ReleaseBinaryCommand extends AppCommand {
 
       const statusCode = response.statusCode;
       if (statusCode >= 400) {
+        debug(`Got error response: ${inspect(response)}`);
         throw statusCode;
       }
       return result;
```
jerietve: Please fix the message below saying "failed to set distribution group"

jerietve: And the other error talking about distribution group

jerietve: This will make tests fail. I tried to do this last week.

PR #687, opened 2019-10-29
Fix build version/number for "publish" ops

### LINKED ISSUE #686: Can't upload DMG
Steps to reproduce:
```
appcenter.cmd distribute groups publish -g <group name> -f <path to DMG>  -r <notes> --build-version <valid version> -a <owner>/<app name> --token <API token>
```

I get the following error:
```
Error: --build-version and --build-number must both be specified when uploading .dmg files
```

If I add `--build-number`, then I get:
```
Error: Command 'distribute groups publish -g X -f Y.dmg -r Z --build-version V -a O/A --token T --build-number N' failed with exception "Unknown argument --build-number"
```

How can I correctly upload a DMG to App Center?
It is certainly doable using the website.

jpeg-ms:
Groups support all file formats (except .aab) and may require build-version and build-number.
Stores don't support any file format that would require version info.

This should fix #686

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -532,7 +532,7 @@
     },
     "abbrev": {
       "version": "1.1.1",
-      "resolved": false,
+      "resolved": "",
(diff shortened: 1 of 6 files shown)


jerietve:
Good fix, quick turnaround!

PR #704, opened 2019-11-19
Retry test run status checks

jonstoneman:
Retry test run status checks

This will allow us to recover from transient network issues

### DIFF
--- a/src/commands/test/lib/state-checker.ts
+++ b/src/commands/test/lib/state-checker.ts
@@ -45,6 +45,11 @@ export class StateChecker {
           throw error;
         }
 
+        if (this.timeIsUp(timeoutSec, startTime, errorRetryWait)) {
+          exitCode = ExitCodes.Timeout;
+          break;
+        }
+
         await out.progress("Status check failed, retrying...", this.delay(1000 * errorRetryWait));
         continue;
       }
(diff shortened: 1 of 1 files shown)


src/commands/test/lib/state-checker.ts
```diff
@@ -28,12 +28,29 @@ export class StateChecker {
...
+      try {
+        state = await out.progress("Checking status...", this.getTestRunState(this.client, this.testRunId));
+      } catch (error) {
+        errorCount++;
+        if (errorCount >= maxErrors) {
+          throw error;
+        }
+
+        await out.progress("Status check failed, retrying...", this.delay(1000 * errorRetryWait));
+        continue;
```
jerietve: This shortcuts the existing timeout behavior. Please make sure the method still exits after `timeoutSec` seconds.

jpeg-ms:
At a quick glance, this looks fine, but was it tested? If yes, then how? (just curious there)

jonstoneman:
I started test runs and disabled wifi - it showed the status check retries occurred and eventually errored correctly. If I re-enabled wifi before the retry limit was hit then the status checks started working again.

I also added temporary code to set timeoutSec to a low number and confirmed that the timeout works with and without wifi disabled.

jonstoneman:
Is there a release planned or could we get one for this change?

jerietve:
@joabalea and @jpeg-ms are planning a release today. You can check the Slack channel for details.

joabalea:
@jonstoneman We just released v2.3.1

jonstoneman:
Thanks!

PR #714, opened 2019-11-28
Handle multiple test_env or app_env test parameters

jonstoneman:
Handle multiple test_env or app_env test parameters and throw an error if multiple parameters of other types are supplied

### DIFF
--- a/src/commands/test/lib/parameters-parser.ts
+++ b/src/commands/test/lib/parameters-parser.ts
@@ -4,7 +4,14 @@ export function parseTestParameters(testParameters: string[]) {
   if (testParameters) {
     testParameters.forEach((p) => {
       const parsedParameter = parseTestParameter(p);
-      result[parsedParameter.key] = parsedParameter.value;
+      if (result[parsedParameter.key] != null && (parsedParameter.key === "test_env" || parsedParameter.key === "app_env")) {
+        const combinedValue = `${result[parsedParameter.key]}|${parsedParameter.value}`;
+        result[parsedParameter.key] = combinedValue;
+      } else if (result[parsedParameter.key] == null) {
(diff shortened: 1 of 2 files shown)


jonstoneman:
It would be nice to unit test this method but I don't know how to go about that and have already spent more time on this than I was meant to. If anyone from the CLI team has the time and inclination to add one then that would be greatly appreciated :)

jpeg-ms:
Can you explain what this does? With an example.

jmoody:
For both `app_env` and `test_env` we expected to be able to use `--test-parameter` like this:

```
  --test-parameter "test_env=USERNAME=clever_user@example.com" \
  --test-parameter "test_env=PASSWORD=pa$$w0rd" \
  ...
  --test-parameter "app_env=VERBOSE_LOGGING=1" \
  --test-parameter "app_env=CONTENT_SERVER=staging.example.com \
```

Currently, the CLI captures only the last `test_env` and `app_env`.  In the example above, Test Cloud only receives `"test_env: PASSWORD=pa$$w0rd"` and `"app_env: CONTENT_SERVER=staging.example.com"`.

Users can bypass this using `|` to define multiple env vars in one --test-parameter

```
  --test-parameter "test_env=USERNAME=clever_user@example.com|PASSWORD=pa$$w0rd" \
  ...
  --test-parameter "app_env=VERBOSE_LOGGING=1|CONTENT_SERVER=staging.example.com \
```

We would prefer to offer both options.

jonstoneman:
Is it OK to merge this?

jpeg-ms:
> Is it OK to merge this?

I will have a look at this today.

jonstoneman:
Thank you!

PR #718, opened 2019-12-06
Add first functional test that uses an account to perform real actions

jerietve:
Update the Azure Pipelines config to download the API token for an account in the integration environment from the key vault. It passes that account to the PowerShell script which invokes Pester. That script logs in the user, and then tests run using that account. A single simple test creates an app, checks it worked and deletes it again.


azure-pipelines.yml
```diff
@@ -51,9 +51,16 @@ jobs:
...
 
+  - task: AzureKeyVault@1
+    inputs:
+      azureSubscription: 'AppCenter - INT'
+      KeyVaultName: 'distributionStoresInt'
+      SecretsFilter: 'ac-cli-account-api-token'
+
   - task: PowerShell@2
     inputs:
       filePath: '$(Build.SourcesDirectory)/test/functional/Main.ps1'
```
yujshin: Isn't the name "Main.ps1" too generic :-P

test/functional/commands/apps/Create.Tests.ps1
```diff
@@ -0,0 +1,24 @@
...
+
+    # Assert
+    $app.displayName | Should -Be $appDisplayName
+    $app.os | Should -Be $appOs
+    $app.platform | Should -Be $appPlatform
+
+    $newNumberOfApps = (appcenter apps list --output json | ConvertFrom-Json).Count
+    $newNumberOfApps | Should -Be ($previousNumberOfApps + 1)
+
+    # Aftermath
```
yujshin: 'Aftermath' -> 'Cleanup'? Can we do that like AfterEach or something like that?

jerietve: Doesn't start with A! ^^
I'll structure the tests better with before(each) and after(each) code once more tests appear (next week).

azure-pipelines.yml
```diff
@@ -51,9 +51,16 @@ jobs:
       testRunTitle: 'Verify unit test result'
     displayName: 'Verify unit test result'
 
+  - task: AzureKeyVault@1
+    inputs:
+      azureSubscription: 'AppCenter - INT'
+      KeyVaultName: 'distributionStoresInt'
```
jpeg-ms: Why is this in code?

test/functional/Invoke-FunctionalTests.ps1
```diff
@@ -1,5 +1,11 @@
 #!/usr/bin/env pwsh
 
+param
+(
+  # The API token used to connect to the App Center integration environment to run the integration tests
```
jpeg-ms: You could use `.SYNOPSIS` and `.PARAMETER` here :)
Also mark as mandatory:
```
[String][Parameter(Mandatory = $true)] $Token
```

Issue #692, opened 2019-11-06
Specify mandatory value for `appcenter distribute release`


fabiendem:
Hello,

Would it be possible to add the option `mandatory` to `appcenter distribute release`?

`appcenter distribute groups publish` doesn't support this neither.

At the moment, it's seems impossible to upload a mandatory build to a group in one shot.

The only solution seems to upload the build to another group, and _then_ add it to the mandatory download group using `appcenter distribute releases add-destination`                                    

```
appcenter help distribute releases add-destination                                                                 

Distribute an existing release to an additional destination

Usage: appcenter distribute releases add-destination -d|--destination <arg> -t|--type <arg> -r|--release-id <arg> [-s|--silent] [-m|--mandatory]
         [-a|--app <arg>]

Options:
    -d|--destination <arg>             The name of the store or group, or the email of the tester
    -t|--type <arg>                    The type of destination: [store, group, tester]
    -r|--release-id <arg>              The ID of the release
    -s|--silent                        If set, do not send a notification to the testers (Not used for stores)
    -m|--mandatory                     Whether the release is mandatory for the testers (Not used for stores)
    -a|--app <arg>                     Specify app in the <ownerName>/<appName> format
```

https://openapi.appcenter.ms/#/distribute/releases_addDistributionGroup seems to provide the option.

It would be convenient to specify if the build is `enabled` or not as well when calling `appcenter distribute release`.
It would allow us to enable it from the AppCenter website manually when we are ready for example.

Let me know if you see some easy workaround 👍 

Cheers!

jerietve:
I don't see a better workaround than the one you found.

The CLI would need a little update to allow for this to work. In the code creating the release, similarly to how it afterward updates the release to set the release notes, some call needs to be added to the API to set the mandatory flag (and notify users, while we're at it).

fabiendem:
Thanks for the reply.
I released that instead of 
> upload the build to another group, and then add it to the mandatory download group using `appcenter distribute releases add-destination`

I can:
```sh
# Upload
`appcenter distribute release --app "Owner/App" --group "TheGroup" [.....]`

# Get the release id from the upload above
release_id=`appcenter distribute releases list --app "Owner/App" | grep ID | head -1 | tr -s ' ' | cut -f2 -d ' '`

# Re-trigger the distribution with the mandatory field on the same group
appcenter distribute releases add-destination --app "Owner/App" --destination "TheGroup" --type group --release-id $release_id --silent --mandatory
````

[...]

PR #723, opened 2019-12-18
Adding --mandatory to distribute-release

yujshin:
#692

### DIFF
--- a/src/commands/distribute/release.ts
+++ b/src/commands/distribute/release.ts
@@ -62,6 +62,10 @@ export default class ReleaseBinaryCommand extends AppCommand {
   @longName("silent")
   public silent: boolean;
 
+  @help("Whether the release is mandatory for the testers (Not used for stores)")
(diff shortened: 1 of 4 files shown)


src/commands/distribute/release.ts
```diff
@@ -62,6 +62,10 @@ export default class ReleaseBinaryCommand extends AppCommand {
   @longName("silent")
   public silent: boolean;
 
+  @help("Whether the release is mandatory for the testers (Not used for stores)")
```
jp-andre: Since this does not take in an argument, maybe reword as "make this release mandatory for testers". One way or another, make it clear that the default is `false`. :)

karmaobserver:
For automatically build with some other services like Bitbucket for example, in "build configuration" section, there is no checkbox to distribute release as mandatory (it distributes as not mandatory version). Is there any way to make it mandatory automatically?

PR #739, opened 2020-02-04
Default `buildGradle["project.ext.react"]` to empty array

PeterKruzlics:
It's likely an issue with our project setup, but the android codepush command fails when running:
`return Array.from(buildGradle["project.ext.react"]).includes("enableHermes: true");` because the `buildGradle["project.ext.react"]` returns a blank value.

I would like to propose defaulting the value to an empty array `[]` before performing the `.includes("enableHermes: true");` check.

Please let me know if there's any other details / tests I can/should add 😅

### DIFF
--- a/src/commands/codepush/lib/react-native-utils.ts
+++ b/src/commands/codepush/lib/react-native-utils.ts
@@ -328,7 +328,7 @@ export function getHermesEnabled(gradleFile: string): boolean {
       throw new Error(`Unable to parse the "${buildGradlePath}" file. Please ensure it is a well-formed Gradle file.`);
     })
     .then((buildGradle: any) => {
-      return Array.from(buildGradle["project.ext.react"]).includes("enableHermes: true");
+      return Array.from(buildGradle["project.ext.react"] || []).includes("enableHermes: true");
     });
 }
 


jp-andre:
Hi! Thanks for the PR.
It would indeed be interesting to figure out why that value is empty in the first place. Do you think you could have a quick look at that?

PeterKruzlics:
Screen Shot 2020-02-05 at 9 08 03 AM

I believe it's likely due to how we have this structured, and typescript's attempt to parse it.

If we'd rather not make this change to the appcenter-cli, I can try and figure out a fix within our app

jp-andre:
> I believe it's likely due to how we have this structured, and typescript's attempt to parse it.
> If we'd rather not make this change to the appcenter-cli, I can try and figure out a fix within our app

To be entirely honest, if there's a way to fix this from within your app, that would be ideal for us.

Otherwise we may need to change the logic in our CLI code to support various structs like:
```gradle
project { ext { react { ... } } }
```
or (like yours)
```gradle
project.ext { react { ... } }
```

Labelling this as a bug for now.

PeterKruzlics:
Going to close this for now 👍

PR #742, opened 2020-02-12
Provide suggested fix in set-current token error

mattjegan:
When trying to run `appcenter apps set-current` with either `--token` or the `APPCENTER_ACCESS_TOKEN` environment variable the user receives an error saying that `set-current` doesn't support that auth mechanism. The fix is to use `MOBILE_CENTER_CURRENT_APP` but the error doesn't mention that. This PR resolves this so users know how to proceed.


LooTKinGz:
New repository files

mattjegan:
Thanks for the merge @anywherepilot

PR #750, opened 2020-03-04
Update help on --Fixture-chunk in Test

Oddj0b:


src/commands/test/lib/help-messages.ts
```diff
@@ -51,11 +51,11 @@ export module Messages {
...
       export const RunLocale = "The system locale for the test run. For example, en_US";
       export const RunLanguage = "Override the language (iOS only) for the test run";
-      export const Fixture = "NUnit fixture / namespace to run. If used with include / exclude the fixture filter is applied first (Can be used multiple times)";
+      export const Fixture = "NUnit fixture / namespace to run. If used with include- / exclude-category the fixture filter is applied first (Can be used multiple times)";
       export const IncludeCategory = "NUnit category to run. (Can be used multiple times)";
       export const ExcludeCategory = "NUnit category to not run. (Can be used multiple times) (When include and/or fixture are combined with exclude, all tests with the included categories are run, except for those also marked with the excluded categories)";
       export const TestChunk = "Run tests in parallel by method";
-      export const FixtureChunk = "Run tests in parallel by fixture";
+      export const FixtureChunk = "Run tests in parallel by fixture, but will ignore include- / exlude-category";
```
phillipleblanc: exclude*

src/commands/test/lib/help-messages.ts
```diff
@@ -51,11 +51,11 @@ export module Messages {
...
       export const RunLocale = "The system locale for the test run. For example, en_US";
       export const RunLanguage = "Override the language (iOS only) for the test run";
-      export const Fixture = "NUnit fixture / namespace to run. If used with include / exclude the fixture filter is applied first (Can be used multiple times)";
+      export const Fixture = "NUnit fixture / namespace to run. If used with include- / exclude-category the fixture filter is applied first (Can be used multiple times)";
       export const IncludeCategory = "NUnit category to run. (Can be used multiple times)";
       export const ExcludeCategory = "NUnit category to not run. (Can be used multiple times) (When include and/or fixture are combined with exclude, all tests with the included categories are run, except for those also marked with the excluded categories)";
       export const TestChunk = "Run tests in parallel by method";
-      export const FixtureChunk = "Run tests in parallel by fixture";
+      export const FixtureChunk = "Run tests in parallel by fixture, but will ignore include- / exclude-category";
```
anywherepilot: Tiny: These two sentences are not in the same tense, but should be. I'd stick with "run tests but ignore". The "but" is not necessary in my opinion.

Oddj0b: "Run tests in parallel by fixture, ignore the use of include- / exclude-category"

Do you think this is better?

anywherepilot: Yes, that's great

PR #751, opened 2020-03-06
[CodePush] Add support for Xcode11

andreidubov:
**Issue:** Starting with Xcode 11 when changing the version of the application, it is saved into the `project.pbxproj` instead of `Info.plist`. And link to `$MARKETING_VERSION` is put in `info.plist`. This does not allow appcenter-cli to make the release a new version, because it cannot get the current version from `Info.plist`. Moreover, customers can change the version for each build configuration (Release or Debug and etc.) individually.

**Solution:** We implemented logic that allows getting the `$MARKETING_VERSION` from the `project.pbxproj` file. Appcenter-cli will take the default version of `Release` build configuration, But if the customer wants to create a release for another Build configuration, he can add an option flag with the Build configuration name to the react-release command.


src/commands/codepush/release-react.ts
```diff
@@ -48,6 +48,12 @@ export default class CodePushReleaseReactCommand extends CodePushReleaseCommandB
   @hasArg
   public plistFilePrefix: string;
 
+  @help("Name of build configuration which specifies the binary version you want to target this release at. For example, \"Debug\" or \"Release\" (iOS only)")
+  @shortName("c")
+  @hasArg
+  @longName("build-configuration-name")
+  public buildConfigurationName: string;
```
jp-andre: Is this a mandatory argument?

jp-andre: Oh I just saw in another line that it defaults to Release. Why not here?

andreidubov: Yes, you're right. I already fixed it. Thanks!

src/commands/codepush/lib/react-native-utils.ts
```diff
@@ -77,7 +87,37 @@ export async function getReactNativeProjectAppVersion(versionSearchParams: Versi
...
+          throw new Error(`Unable to find the build configuration with "${buildConfigurationName}" name.`);
+        }
+
+        const marketingVersion = Object.values(configsObj).find((c) => c.buildSettings["MARKETING_VERSION"]).buildSettings.MARKETING_VERSION;
+        if (!isValidVersion(marketingVersion)) {
+          throw new Error(`The "MARKETING_VERSION" key in the "${resolvedPbxprojFile}" file needs to specify a valid semver string, containing both a major and minor version (e.g. 1.3.2, 1.1).`);
+        }
+        out.text(`Using the target binary version value "${marketingVersion}" from "${resolvedPbxprojFile}".\n`);
+
+        return Promise.resolve(marketingVersion);
```
pfleidi: I think we can skip the `Promise.resolve()` here since the method is `async` already.

pfleidi:
@jp-andre We'll schedule a feature to add missing tests for all of the release commands in the next weeks. Right now, there are no tests for most of the `release-react` functionality and adding them would block the xcode 11 customers for a while since we would have to add a lot of infrastructure for these tests. We'll make sure everything works by validating the code manually and add tests once we have enough bandwidth.

Does that make sense to you?

jp-andre:
> @jp-andre We'll schedule a feature to add missing tests for all of the release commands in the next weeks. Right now, there are no tests for most of the `release-react` functionality and adding them would block the xcode 11 customers for a while since we would have to add a lot of infrastructure for these tests. We'll make sure everything works by validating the code manually and add tests once we have enough bandwidth.
> 
> Does that make sense to you?

Yes, considering the situation this is perfectly understandable. I'll work on a release.

PR #757, opened 2020-03-17
Update dependencies

anywherepilot:
Update dependencies and lock their versions. Brief overview of changes below. Internal quality updates (moving files about, updating doc, etc.) not mentioned.

Major updates:

- mocha: drops old Node support. Changes to the flow when conditionally skipping tests. Changes to command line options which we don't use.
- rimraf: drops old Node support.
- semver: drops old Node support.
- strip-ansi: drops old Node support. Now forces uses of `require(strip-ansi)` syntax which we already use.
- ts-node: bug fixes and minor features. Latest release is 8.7.0 but it's not on npmjs yet even though it was created two weeks ago.
- tslint: some rule changes that don't affect our codebase. We need to use ESLint instead.
- update-notifier: removes features we don't use, fixes security vulnerability.
- uuid: removes features we don't use and old Node support.
- which: drops old Node support.

Minor updates:

- fast-xml-parser: bug fixes.
- typescript: lots of new stuff.
- xmldom: drops old Node support.

Patch updates:

- mocha-junit-reporter
- p-limit
- request
- rxjs
- unzipper
- xml2js

Not updated:

- autorest: would need to test the tool, i expect changes.
- chalk: some breaking changes.
- date-fns: looks like some work.
- inquirer: breaking changes.
- mkdirp: full rewrite, now returns a promise. Needs some work. Probably makes parts of promisfied-fs.ts obsolete.
- apcenter-file-upload-client, gulp-typescript, nock, sinon, webpack-cli: ran out of boxed time.


jp-andre:
What is "old Node"?

anywherepilot:
> What is "old Node"?

4, 6, in some cases 8.

jp-andre:
> Would this be a patch version or minor version?

I'd say micro/patch since there is no functional change.

yujincat:
I guess the ac-cli version will be updated in a separate PR..?

anywherepilot:
I am planning to make this a patch release. The not yet updated packages should all be updated as well but will take time. I'll handle them later this week load permitting (it's looking good at the moment).

jorgengranseth:
@anywherepilot when will this (version `2.3.5`?) be available on npm?

anywherepilot:
Hi @jorgengranseth . We'll have to update all dependencies before we can release, since there's no point updating the security issue in one place if it's still present in others. I expect a release later this week.

jp-andre:
See #762

PR #758, opened 2020-03-19
Version to 2.3.5

anywherepilot:
Bump patch version for release of dependency updates


anywherepilot:
Well, great. Turns out nock, mkdirp, unzipper, mocha, tslint AND gulp use minimist and we'll have to update them all.. Doing this release without updating them is pointless, because the only other thing this release does is update the rest of the dependencies not on that list.

anywherepilot:
I'm confused this came up now, rather than with the previous PR. Anyway, postponing till next week.

Richienb:
@anywherepilot Many people are reporting deprecation messages from `request`. Can we replace it with `got` at some point?

anywherepilot:
Hi @Richienb . It's the first I've heard of it. Could you open an issue so we can track it effectively? I was currently just working on upgrading dependencies, not replacing them. If something is deprecated then it should be replaced of course.

PR #760, opened 2020-03-25
Fix build URLs with slashes in branch names. Fixes #755

### LINKED ISSUE #755: Build branches list gives invalid URL when branch has a slash (/) in it
Eg the command: `appcenter build branches list --output json --app org/app`

Will return a URL to a 404 page when the branch built has a slash (/) in it.

It looks like the forward slash in the branchname is not correctly being escaped: (eg / should be %2F)

jp-andre:
This fixes #755.


jp-andre:
Waiting on #762

jp-andre:
> Please review my addition before merging
That change was somewhere on my local machine. Apparently it was lost in git :) Good catch!

anywherepilot:
I looked into https://www.npmjs.com/package/filenamify . Just replacing slashes is not enough on Windows.

PR #762, opened 2020-03-25
Update a few deps, fix issues with security audit

jp-andre:


src/commands/distribute/groups/download.ts
```diff
@@ -148,37 +148,38 @@ export default class DownloadBinaryFromDistributionGroupCommand extends AppComma
...
+      }
     } else {
       // using current working directory by default
       return Promise.resolve(cwd());
     }
   }
 
-  private getFileFullPath(passedFileName: string, directoryPath: string, appName: string, downloadUrl: string): string {
+  private getFilenameFromDownloadUrl(downloadUrl: string): string {
+    const filename = Url.parse(downloadUrl).pathname?.split("/").slice(-1)[0];
```
anywherepilot: Won't this select just a single character? The `slice(-1)` selects the last string from the array of strings, and then `[0]` selects the first character of that string.

src/commands/distribute/groups/download.ts
```diff
@@ -148,37 +148,38 @@ export default class DownloadBinaryFromDistributionGroupCommand extends AppComma
...
+    debug(`Got filename from URL: ${filename}`);
+    return filename;
+  }
+
+  private getFileFullPath(passedFileName: string, directoryPath: string, downloadUrl: string): string {
```
anywherepilot: I'm sure these changes are fine, but how are they related to the dependency change?

typings/index.d.ts
```diff
@@ -1,4 +1,3 @@
 /// <reference path="modules/azure-storage/index.d.ts" />
```
anywherepilot: Why are/were files like this in the repo?

src/commands/distribute/groups/download.ts
anywherepilot: Ah no I'm stupid ^^ Slice returns another array of course. Can't you take `[-1]` from the array? Is that Ruby?

jp-andre: You can't do [-1]. I asked StackOverflow for the one-liner.

src/commands/distribute/groups/download.ts
jp-andre: Good catch! This is an overdue fix for the new ACFUS style URLs.

PR #765, opened 2020-03-26
[WIP] Replace deprecated "request" with "got"

jp-andre:
This is currently work in progress.
Got v10 has some issues with nock which required some black magic (convert call like `request(url, options)` to `request(optionsWithUrl)`).

This comes after #762


Richienb:
@jp-andre Got comes bundles with typings - `@types/got` should be a stub.

sirsavary:
I'm concerned that this PR changes a lot more than replacing `request` with `got`. It adds an extra dependency and modifies the usage of `mkdirp` throughout the codebase.

Also, what's the reasoning behind selecting `got` over other choices listed on https://github.com/request/request/issues/3143? I'm wondering if a smaller library would be better suited, seeing as how `request` was only used in a handful of places.

Richienb:
@sirsavary Technically, we could also use `node-fetch` instead as a more light-weight alternative but the motivation for `got` is its ease of use.

jp-andre:
Honestly I went with got for two simple reasons: 1. it's massively used so we can expect long-term support, and 2. it's said to be easy to use (well that was not true at all with regards to nock & unit testing).
This PR is a draft. All changes unrelated to request are likely already merged from the PR I mentioned earlier (#762).
We could use node-fetch or any other library provided it supports a minimum set of features that we need (and honestly I don't know exactly what we need at this point) and would require minimal changes in our code.

jp-andre:
Abandoning old PR.

PR #769, opened 2020-03-31
[WIP] Update autorest

jp-andre:
ATTENTION This is Work in progress
Latest autorest is unable to generate the client API with the current contents of openapi.appcenter.ms
Older autorest (2.x) is still able to generate an API but:
- requires 16Gb of memory
- changes the client API in a way that breaks compilation and would require extra changes


jp-andre:
Abandoning old PR.

Issue #770, opened 2020-03-31
Switch to upstream eslint


jp-andre:
Remove custom tslint rules, and use upstream eslint. Use the most standard packages.

ajaykn:
@anywherepilot closing the issue, as this is merged.

PR #778, opened 2020-04-01
Onboard ESLint and prettier

anywherepilot:
Remove TSLint, since it was deprecated last year.


anywherepilot:
Please review each commit by itself, especially the one called "Ignore most and fix some ESLint errors".

anywherepilot:
This addresses #770

gulpfile.js
```diff
@@ -1,97 +1,84 @@
...
-  rimraf('dist', done);
+gulp.task("clean", function (done) {
+  rimraf("dist", done);
 });
 
-gulp.task('build-ts', function () {
-  let tsResult = gulp.src(['src/**/*.ts', 'typings/**/*.d.ts'])
-    .pipe(tsProject());
+gulp.task("build-ts", function () {
+  const tsResult = gulp.src(["src/**/*.ts", "typings/**/*.d.ts"]).pipe(tsProject());
```
jp-andre, line 21: we can probably remove typings from here

.prettierignore
```diff
@@ -0,0 +1,2 @@
+/dist
+/src/util/apis/generated
```
jp-andre, line 2: can we have a newline here? (if prettier accepts)

gulpfile.js
```diff
@@ -1,97 +1,84 @@
...
-const autocompleteTree = require('./scripts/autocomplete-tree');
+const clean = require("gulp-clean");
+const gulp = require("gulp");
+const rimraf = require("rimraf");
+const sourcemaps = require("gulp-sourcemaps");
+const ts = require("gulp-typescript");
+const autocompleteTree = require("./scripts/autocomplete-tree");
 
-let tsProject = ts.createProject('tsconfig.json');
+const tsProject = ts.createProject("tsconfig.json");
```
jp-andre, line 10: maybe those can use import syntax? just wondering

scripts/all-command-help.js
```diff
@@ -4,99 +4,99 @@
...
+const os = require("os");
+const path = require("path");
+const util = require("util");
+const Table = require("cli-table3");
+const {
+  getClassHelpText,
+  getOptionsDescription,
+  getPositionalOptionsDescription,
+} = require("../dist/util/commandline/option-decorators");
+const { getOptionsForTwoColumnTableWithNoBorders } = require("../dist/util/interaction/out");
```
jp-andre, line 20: maybe import here as well?

scripts/all-command-help.js
```diff
@@ -4,99 +4,99 @@
...
 function checkStats(dir, fileName, checker) {
   try {
-    let s = fs.statSync(path.join(dir, fileName))
-    return checker(s)
-  }
-  catch (err) {
-    return false
+    let s = fs.statSync(path.join(dir, fileName));
```
jp-andre, line 23: looks like this should be const -- why no linting error?

package.json
```diff
@@ -17,16 +17,18 @@
   },
   "preferGlobal": true,
   "scripts": {
-    "build": "npm run tslint && gulp build:raw",
-    "tslint": "tslint --project .",
-    "test": "npm run tslint && mocha --require ts-node/register ./test/**/*-test.[tj]s",
+    "build": "npm run lint && gulp build:raw",
+    "lint": "eslint \"src/**/*.ts\" \"test/**/*.ts\"",
```
jp-andre, line 21: add scripts and bin to that (if possible of course)

anywherepilot:
I'm curious whether you think the changes from for loops to for-of loops are safe. One of them actually caused test failures saying the object wasn't iterable. I don't know how it does pass with a normal for loop. See last commit.

.prettierignore
anywherepilot: Oh yeah of course. Prettier actually enforces that, but doesn't run on this file type.

anywherepilot: Nope it does not enforce that... I must be thinking of some other tool. But I'll add it, no problem.

jp-andre:
> I'm curious whether you think the changes from for loops to for-of loops are safe. One of them actually caused test failures saying the object wasn't iterable. I don't know how it does pass with a normal for loop. See last commit.

Yeah I saw those changes. A quick search shows that the loop will not respect the order of values.
I would rather use indices as there's a risk in changing that.

anywherepilot:
I copied the rules from the common App Center package. We can't import that package here, because it is in the internal feed.

gulpfile.js
anywherepilot: Good find. I also removed it from the doc.

.prettierignore
anywherepilot: Done!

gulpfile.js
anywherepilot: I assume you're referring to the `require` statements above. I read up on this syntax a bit. It should work, but doesn't. I'm getting unexpected token errors on the asterisk. Maybe we're not using ES6. I'll leave this for this PR.

scripts/all-command-help.js
anywherepilot: Not touching the import syntax of .js files.

scripts/all-command-help.js
anywherepilot: We never linted JavaScript before. I enabled it and applied fixes.

package.json
anywherepilot: Added.

PR #797, opened 2020-04-14
gitIgnoring visual code ide files

BalajiJBcs:
Add the visual code IDE files to gitIgnore, to prevent tracking of the IDE settings of different developers.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,6 +3,7 @@
 
 # IDE files
 .idea
+.vscode
 
 # Logs
 logs


ajaykn:
HI @BalajiJBcs, Really appreciate your efforts in making this product better.

Reason we are tracking the files in .vscode folder, for instance 
- launch.json needs to be tracked so that we can have launch and debug actions thats used for local debugging by contributors.
- settings.json, we need this is to make sure everyone follows the same settings 
for example: we use tabSize: 2, across this project and we dont want the contributors use different settings.
Anyways, if this gets overridden by mistake, we will handle while reviewing the PR's.

So I am thinking not to go ahead with this PR.
Do you have any suggestions. ?

BalajiJBcs:
Ya sure, The issue with this approach if some one playing around the vs code setting it will be reflected in git logs. When we do git add *, it add this file. As most people will do rather than using git add filename. i have seen multiple issues in such case where unwanted files get added to the git logs. It becomes much cumber some to remove. As Some commits of this  kind of files in middle of the cherry pick. To solve this issues I usually recommend to remove the .vscode. But on top of all. It seems you have valid reason why we need to do this. I understand the requirement why it is in place and removed my pull request. Thanks for your reply @ajaykn

PR #798, opened 2020-04-14
Juggling check  for handling null and undefined

BalajiJBcs:
Check for null and undefined using the juggling method. In which both null and undefined handle in a single hit.


ajaykn:
@BalajiJBcs Thanks for raising the PR.
But the file you are changing is an auto generated one by [Autorest](https://www.npmjs.com/package/autorest).
We have swagger enddoints and we run auto rest to generate this lib code.
So the manual changes to this file are overridden once we run the auto rest in future.

I am not sure, if we want to go ahead with this PR.
Do you have any concerns ?

BalajiJBcs:
Have closed the request based on @ajaykn  comments

PR #806, opened 2020-04-16
Disable telemetry and the telemetry question on the `telemetry` command.

### LINKED ISSUE #485: Disable telemetry check when running `appcenter telemetry`
When running
```
> appcenter telemetry off
```
I get the following prompt
```
App Center CLI would like to collect data about how users use CLI commands and some problems they
encounter. Participation is voluntary and when you choose to participate your device automatically sends
information to Microsoft about how you use App Center CLI.
For more information, please see our privacy policy at https://aka.ms/mobilecenterprivacy

? Enable telemetry?
```

This seems like counter intuitive behaviour and this check should should not happen when running the `telemetry` command

## Steps to reproduce
Run
```
rm ~/.appcenter-cli/telemetryEnabled.json
```
then run any `appcenter telemetry` command.

## Workaround
For now, I am adding `--disable-telemetry` to the commands. EG:
```
appcenter telemetry off --disable-telemetry
```

anywherepilot:
Resolves #485

### DIFF
--- a/src/util/commandline/command.ts
+++ b/src/util/commandline/command.ts
@@ -137,7 +137,12 @@ export class Command {
         return Promise.resolve(Result.failure(Result.ErrorCodes.InvalidParameter, `Unknown output format ${this.format}`));
       }
     }
-    this.clientFactory = createAppCenterClient(this.command, await telemetryIsEnabled(this.disableTelemetry));
+
+    this.clientFactory = createAppCenterClient(
+      this.command,
+      await telemetryIsEnabled(this.disableTelemetry || this.command[0] === "telemetry")
+    );
+
     return this.runNoClient();
   }
 
--- a/src/util/commandline/command.ts
+++ b/src/util/commandline/command.ts
@@ -137,9 +137,7 @@ export class Command {
         return Promise.resolve(Result.failure(Result.ErrorCodes.InvalidParameter, `Unknown output format ${this.format}`));
       }
     }
-
     this.clientFactory = createAppCenterClient(this.command, await telemetryIsEnabled(this.disableTelemetry));
-
     return this.runNoClient();
   }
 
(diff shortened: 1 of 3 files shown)


src/util/commandline/command.ts
```diff
@@ -137,7 +137,12 @@ export class Command {
         return Promise.resolve(Result.failure(Result.ErrorCodes.InvalidParameter, `Unknown output format ${this.format}`));
       }
     }
-    this.clientFactory = createAppCenterClient(this.command, await telemetryIsEnabled(this.disableTelemetry));
+
+    this.clientFactory = createAppCenterClient(
+      this.command,
+      await telemetryIsEnabled(this.disableTelemetry || this.command[0] === "telemetry")
```
jp-andre, line 143: If I understand correctly this will always disable telemetry on that command, even if telemetry was previously enabled (i.e. false vs. undefined). It probably does not really matter but I'd like to make sure that was the intent.

anywherepilot: It will not emit telemetry when calling this command, regardless of the effect of the command itself (enabling or disabling telemetry for other commands). Alternative solution is to put `disableTelemetry = true` inside the constructors of the `telemetry` `on` and `off` commands.

anywherepilot: And that is the intent. If you're performing an operation enabling or disabling telemetry, you don't then during that want to be asked a question about it, performing the operation twice and confusing you.

jp-andre: I think this approach is fine. It's simple.

anywherepilot: I like the other solution a bit better since it doesn't mess with the main `Command` class so I'll update the PR to it. Same effect :)

PR #808, opened 2020-04-16
Allow changing whether a distribution group is public.

### LINKED ISSUE #642: Set public distribute group access
We currently use HockeyApp to distribute test builds of our application to testers. These testers don't have a Microsoft account and forcing them to create one before testing would result in many of them just not bothering.

There is an option in the web platform to enable public access: 
image

The issue is that this requires us to manually do it which doesn't work well with our current build processes where everything is automated. 

Is it possible that this feature can be incorporated into the CLI? It's the main blocker stopping us from moving over to App Center from HockeyApp. Something like _appcenter distribute groups update --set-public true/false_ would be perfect.

anywherepilot:
Add `public` and `private` switches to `distribute groups update`. These allow making a group public or the opposite.

Resolves #642.

### DIFF
--- a/src/commands/distribute/groups/update.ts
+++ b/src/commands/distribute/groups/update.ts
@@ -58,11 +58,14 @@ export default class UpdateDistributionGroupCommand extends AppCommand {
   @hasArg
   public testersToDeleteListFile: string;
 
-  @help("Whether the distribution group is public (allowing anyone to download the releases)")
+  @help("Make the distribution group public (allowing anyone to download the releases)")
   @shortName("p")
   @longName("public")
-  @hasArg
-  public isPublic: string;
+  public makePublic: boolean;
+
+  @help("Make the distribution group private (allowing only members to download the releases)")
+  @longName("no-public")
+  public makePrivate: boolean;
 
   public async run(client: AppCenterClient): Promise<CommandResult> {
     const app = this.app;
@@ -111,9 +114,14 @@ export default class UpdateDistributionGroupCommand extends AppCommand {
       currentGroupName = this.distributionGroup;
     }
 
-    if (!_.isNil(this.isPublic)) {
-      debug("Setting distribution group public status to " + this.isPublic);
-      options.isPublic = JSON.parse(this.isPublic);
+    if (this.makePublic) {
+      debug("Setting distribution group public status to true");
+      options.isPublic = true;
+    }
+
+    if (this.makePrivate) {
+      debug("Setting distribution group public status to false");
+      options.isPublic = false;
     }
 
     if (!_.isNil(options.name) || !_.isNil(options.isPublic)) {
@@ -136,7 +144,8 @@ export default class UpdateDistributionGroupCommand extends AppCommand {
       _.isNil(this.testersToAddListFile) &&
       _.isNil(this.testersToDelete) &&
       _.isNil(this.testersToDeleteListFile) &&
-      _.isNil(this.isPublic)
+      _.isNil(this.makePublic) &&
+      _.isNil(this.makePrivate)
     ) {
       throw failure(ErrorCodes.InvalidParameter, "nothing to update");
     }
@@ -146,8 +155,8 @@ export default class UpdateDistributionGroupCommand extends AppCommand {
     if (!_.isNil(this.testersToDelete) && !_.isNil(this.testersToDeleteListFile)) {
       throw failure(ErrorCodes.InvalidParameter, "parameters 'delete-testers' and 'delete-testers-file' are mutually exclusive");
     }
-    if (!_.isNil(this.isPublic) && this.isPublic.toLowerCase() !== "true" && this.isPublic.toLowerCase() !== "false") {
-      throw failure(ErrorCodes.InvalidParameter, "boolean parameter 'is-public' must be either 'true' or 'false'");
+    if (this.makePublic && this.makePrivate) {
+      throw failure(ErrorCodes.InvalidParameter, "parameters 'public' and 'no-public' are mutually exclusive");
     }
   }
 
--- a/src/commands/distribute/groups/update.ts
+++ b/src/commands/distribute/groups/update.ts
@@ -58,13 +58,13 @@ export default class UpdateDistributionGroupCommand extends AppCommand {
   @hasArg
   public testersToDeleteListFile: string;
 
-  @help("Make the distribution group public (allowing anyone to download the releases)")
+  @help("Make the distribution group public (allowing anyone to download the releases). Don't use with opposite --private.")
   @shortName("p")
   @longName("public")
   public makePublic: boolean;
 
-  @help("Make the distribution group private (allowing only members to download the releases)")
-  @longName("no-public")
+  @help("Make the distribution group private (allowing only members to download the releases). Don't use with opposite --public.")
+  @longName("private")
   public makePrivate: boolean;
 
   public async run(client: AppCenterClient): Promise<CommandResult> {
--- a/src/commands/distribute/groups/update.ts
+++ b/src/commands/distribute/groups/update.ts
@@ -60,7 +60,7 @@ export default class UpdateDistributionGroupCommand extends AppCommand {
 
   @help("Whether the distribution group is public (allowing anyone to download the releases)")
   @shortName("p")
-  @longName("isPublic")
+  @longName("public")
   @hasArg
   public isPublic: string;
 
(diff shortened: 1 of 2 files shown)


src/commands/distribute/groups/update.ts
```diff
@@ -58,6 +58,12 @@ export default class UpdateDistributionGroupCommand extends AppCommand {
   @hasArg
   public testersToDeleteListFile: string;
 
+  @help("Whether the distribution group is public (allowing anyone to download the releases)")
+  @shortName("p")
+  @longName("isPublic")
+  @hasArg
+  public isPublic: string;
```
jp-andre, line 65: `public` or maybe `is-public` though the `is` is not relevant.
Do we natively support bool here by any chance?

jp-andre: Also it looks like this should act like a switch, where `-p` (or even `--public`) alone should set the value to `true`.

anywherepilot: I considered that, but then how do you set it to not public? With another switch?

jp-andre: `--public=false`

jp-andre: we apparently have support for booleans in the base class, let's see what it can do, and whether we need to expand it for this (which would not be worth the effort)

jp-andre: Alternatively, a common convention is to prefix with `no` as in `--no-public` (no short form here)

anywherepilot: We're using names like `isMandatory` and `isDisabled` in other places, but I can rename it to just `public`. We also have precedent for that. It's a mix.

This is not a flag, it's a setter. It's setting a value to true or false, not on the command but on the distribution group it's modifying.

Problem with making it bool is that it will need to default to false, and we'll no longer be able to check whether it is passed or not. TypeScript doesn't support nullable booleans. We can't work with that, because we can't just go set the value based on that information. Since it's a setter, not a flag.

If providing a string value for a boolean is really unwanted, we can have two parameters called `makePublic` and `makePrivate` or something. I'm not sure that looks nicer.

jp-andre: > We're using names like `isMandatory` and `isDisabled` in other places, but I can rename it to just `public`. We also have precedent for that. It's a mix.

Sorry, but that's not true. From the code right there:
```
  @help("Specifies whether this release should be considered mandatory. (Putting -m flag means mandatory)")
  @shortName("m")
  @longName("mandatory")
  public isMandatory: boolean;
```

The option is called `mandatory` and it's a boolean switch. Specifically for casing, the only mistake I could find was `displayKeys` (we may want to consider fixing that in another PR, it should be `display-keys`).

> This is not a flag, it's a setter. It's setting a value to true or false, not on the command but on the distribution group it's modifying.

I understand that. There are ways to make this work. Either the option could also accept an optional argument, or we could use something like `--no-public` or `--private` as you proposed.

> Problem with making it bool is that it will need to default to false, and we'll no longer be able to check whether it is passed or not. TypeScript doesn't support nullable booleans. We can't work with that, because we can't just go set the value based on that information. Since it's a setter, not a flag.

The default value of a boolean field is "the field does not even exist", and "undefined" is perfectly acceptable for a boolean in TS. Yay.

The only question is how much effort is it to have an optional argument to a boolean switch, not whether it's possible or impossible.

test/commands/distribute/groups/update-test.ts
```diff
@@ -5,7 +5,7 @@ import * as _ from "lodash";
 import UpdateDistributionGroupCommand from "../../../../src/commands/distribute/groups/update";
 import { CommandArgs, CommandResult } from "../../../../src/util/commandline";
 
-describe("distribute groups update command", () => {
+describe.only("distribute groups update command", () => {
```
yujincat, line 8: We need to remove this `only`

src/commands/distribute/groups/update.ts
```diff
@@ -91,16 +97,27 @@ export default class UpdateDistributionGroupCommand extends AppCommand {
...
+      currentGroupName = this.newDistributionGroupName;
     } else {
       currentGroupName = this.distributionGroup;
     }
 
-    if (deletedTestersEmails.length !== testersToDeleteEmails.length || addedTestersEmails.length !== testersToAddEmails.length) {
-      out.text("Updating the list of testers was partially successful");
+    if (!_.isNil(this.isPublic)) {
+      debug("Setting distribution group public status to " + this.isPublic);
+      options.isPublic = JSON.parse(this.isPublic);
```
yujincat, line 116: How does it work when `isPublic` is unexpected value? I tested in the console with "TRUE" and it does not work.

src/commands/distribute/groups/update.ts
yujincat: One more vote for having it as `boolean` using `undefined`. Otherwise we should handle the string to be a boolean value and that sound like more work.

anywherepilot: You are right on the naming. I was looking at the variable names, not the argument names.

The way booleans work in the CLI is by not setting the `hasArg` attribute on a property. It then becomes a boolean (regardless of what you say it should be in the property definition). The value is `false` when not provided. There is no `undefined`. Unless we add support for an optional parameter, which would be difficult*, there is mathematically speaking no difference between `false` and `undefined`. It's just the "not true" option, and cannot be used. So we need two switches.

* It would require defining an actual switch type parameter with a new syntax for taking an optional `false` argument.

There is no other example that I know of in the CLI that sets a boolean as obviously as this one, so we have to make a new decision. You and @yujincat are in favor of having two switches (or against a parameter taking a "true"/"false" string), and I don't have an opinion. So let's go with the two switches. I'll call them `--public` and `--no-public` as you suggested.

This is a good discussion btw, I'm glad we're having it :)
Things may be built on top of this in the future.

src/commands/distribute/groups/update.ts
```diff
@@ -91,16 +100,32 @@ export default class UpdateDistributionGroupCommand extends AppCommand {
...
+      options.isPublic = true;
+    }
+
+    if (this.makePrivate) {
+      debug("Setting distribution group public status to false");
+      options.isPublic = false;
+    }
+
+    if (!_.isNil(options.name) || !_.isNil(options.isPublic)) {
+      await this.updateDistributionGroup(client, app, options);
```
yujincat, line 128: I am curious why option validation (like `this.makePublic && this.makePrivate`) is after the `updateDistributionGroup` calls :-P How does that work?

src/commands/distribute/groups/update.ts
```diff
@@ -91,16 +100,32 @@ export default class UpdateDistributionGroupCommand extends AppCommand {
...
-      out.text("Updating the list of testers was partially successful");
+    if (this.makePublic) {
+      debug("Setting distribution group public status to true");
+      options.isPublic = true;
+    }
+
+    if (this.makePrivate) {
+      debug("Setting distribution group public status to false");
+      options.isPublic = false;
+    }
```
jp-andre, line 125: Let's make sure only one of the two is used at once.

anywherepilot: That's in the parameter validation

src/commands/distribute/groups/update.ts
anywherepilot: Can you clarify your question? The call to `validateParameters` is right at the top of the `run` method.

anywherepilot: You might be confused by the lack of context provided by the GitHub changes view :(

yujincat: You are right. I was confused by the unfolded code parts... 😭

test/commands/distribute/groups/update-test.ts
```diff
@@ -67,12 +126,14 @@ describe("distribute groups update command", () => {
...
+    if (abortScope) {
+      expect(abortScope.isDone()).to.eql(false, "Unexpected requests were made");
+    }
     executionScope.done(); // All normal API calls are executed
   }
 
-  function testCommandFailure(executionScope: Nock.Scope, abortScope?: Nock.Scope) {
-    expect(abortScope.isDone()).to.eql(false, "Unexpected requests were made");
+  function testCommandFailure(executionScope: Nock.Scope, skippedScope?: Nock.Scope) {
+    expect(skippedScope.isDone()).to.eql(false, "Unexpected requests were made");
```
yujincat, line 136: Why doesn't it have null check like `testCommandSuccess`?

anywherepilot: A better question would be why is this method copied into every test class 😛 
I didn't need to fix this one to make my code work.

PR #811, opened 2020-04-17
Fix react-native build with hermes enabled on Windows

### LINKED ISSUE #693: "release-react" fails when using RN 0.61+ with Hermes enabled
- `appcenter-cli` 2.3.0
- `react-native` 0.61.4

Builds on Windows fail with error `spawn node_modules\hermesvm\win64-bin\hermes ENOENT`.

Part of the problem is that `node_modules/hermesvm` has been replaced with `node_modules/hermes-engine` starting from RN 0.61. #653 and #654 partially solve the issue, but do not account for the fact that the Hermes executable is `hermes.exe` and not `hermes` on Windows systems, meaning this: 

https://github.com/microsoft/appcenter-cli/blob/bf105d1a7dbec788909b3b35f1c6fa283a31314f/src/commands/codepush/lib/react-native-utils.ts#L354-L357

always fails, which in turn makes AppCenter fall back on `node_modules/hermesvm`, which no longer exists.

Shywim:
This should fixes #693.

For now, we can't deploy a react-native release on Windows with hermes enabled because the script looks for an executable file without a `.exe` file extension, whereas on Windows, we need a `.exe` extension, throwing an ENOENT error.

image

Full error:
```
Error: spawn node_modules\hermesvm\win64-bin\hermes ENOENT
    at Process.ChildProcess._handle.onexit (internal/child_process.js:264:19)
    at onErrorNT (internal/child_process.js:456:16)
    at processTicksAndRejections (internal/process/task_queues.js:77:11)
Emitted 'error' event at:
    at Process.ChildProcess._handle.onexit (internal/child_process.js:270:12)
    at onErrorNT (internal/child_process.js:456:16)
    at processTicksAndRejections (internal/process/task_queues.js:77:11) {
  errno: 'ENOENT',
  code: 'ENOENT',
  syscall: 'spawn node_modules\\hermesvm\\win64-bin\\hermes',
  path: 'node_modules\\hermesvm\\win64-bin\\hermes',
  spawnargs: [
    '-emit-binary',
    '-out',
    'C:\\Users\\shywi\\AppData\\Local\\Temp\\code-push2020317-23456-1rdxgc2.ndt9\\CodePush\\index.android.bundle.hbc',      'C:\\Users\\shywi\\AppData\\Local\\Temp\\code-push2020317-23456-1rdxgc2.ndt9\\CodePush\\index.android.bundle',          '-w'
  ]
}
```

This PR add a function to look for platform specific executable file extension.

### DIFF
--- a/src/commands/codepush/lib/react-native-utils.ts
+++ b/src/commands/codepush/lib/react-native-utils.ts
@@ -438,6 +438,15 @@ function getHermesOSBin(): string {
   }
 }
 
+function getHermesOSExe(): string {
+  switch (process.platform) {
+    case "win32":
+      return ".exe";
+    default:
+      return "";
+  }
+}
+
 function getHermesCommand(): string {
   const fileExists = (file: string): boolean => {
     try {
@@ -447,7 +456,7 @@ function getHermesCommand(): string {
     }
   };
   // assume if hermes-engine exists it should be used instead of hermesvm
-  const hermesEngine = path.join("node_modules", "hermes-engine", getHermesOSBin(), "hermes");
+  const hermesEngine = path.join("node_modules", "hermes-engine", getHermesOSBin(), `hermes${getHermesOSExe()}`);
   if (fileExists(hermesEngine)) {
     return hermesEngine;
   }
--- a/src/commands/codepush/lib/react-native-utils.ts
+++ b/src/commands/codepush/lib/react-native-utils.ts
@@ -441,9 +441,9 @@ function getHermesOSBin(): string {
 function getHermesOSExe(): string {
   switch (process.platform) {
     case "win32":
-      return ".exe";
+      return "hermes.exe";
     default:
-      return "";
+      return "hermes";
   }
 }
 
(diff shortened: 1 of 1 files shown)


src/commands/codepush/lib/react-native-utils.ts
```diff
@@ -447,7 +456,7 @@ function getHermesCommand(): string {
     }
   };
   // assume if hermes-engine exists it should be used instead of hermesvm
-  const hermesEngine = path.join("node_modules", "hermes-engine", getHermesOSBin(), "hermes");
+  const hermesEngine = path.join("node_modules", "hermes-engine", getHermesOSBin(), `hermes${getHermesOSExe()}`);
```
anywherepilot, line 459: Let's have the method return the whole name, rather than just the extension. In general this change looks fine. We can release this quickly.

anywherepilot:
Thank you very much for figuring out the problem and solving it @Shywim. I agree this solves #693. We'll release a new version with this code so you and everyone can benefit.

PR #832, opened 2020-04-24
Run autorest with 16Gb

jp-andre:
Here's a first PR that actually solves running Autorest (version 2) using 16Gb of memory for node.
Tested with node 10.
A follow-up PR will update the client APIs.

### DIFF
--- a/package.json
+++ b/package.json
@@ -25,7 +25,7 @@
     "watch-test": "mocha --require ts-node/register ./test/**/*-test.[tj]s --watch",
     "autorest:clean": "rimraf src/util/apis/generated",
     "autorest": "npm run autorest:clean && node scripts/autorest.js",
-    "autorest:gen": "node --max-old-space-size=16384 node_modules/.bin/autorest --debug swagger/readme.md",
+    "autorest:gen": "node --max-old-space-size=16384 node_modules/autorest/dist/app.js --debug swagger/readme.md",
     "clean": "gulp clean",
     "prepublishOnly": "npm run lint && gulp prepublish",
     "download-swagger": "gulp download-swagger",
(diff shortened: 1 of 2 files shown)


anywherepilot:
Please hold off while I figure out how to make this work on Windows. There's no `/node_modules/bin` path, just `/node_modules/.bin`.

package.json
```diff
@@ -25,6 +25,7 @@
     "watch-test": "mocha --require ts-node/register ./test/**/*-test.[tj]s --watch",
     "autorest:clean": "rimraf src/util/apis/generated",
     "autorest": "npm run autorest:clean && node scripts/autorest.js",
+    "autorest:gen": "node --max-old-space-size=16384 node_modules/bin/autorest --debug swagger/readme.md",
```
anywherepilot, line 28: Does this work for you? Both on my Windows and macOS machines I have a .bin folder in node_modules instead of a bin one, which contains autorest. I think this should be `node_modules/.bin/autorest` just like in the autorest.js file. It still won't work on Windows then because there it's called autorest.cmd instead.
Maybe use a script just like the `autorest` command above?

jp-andre: Ooops, bad copy & paste.
Is the `.cmd` file actually javascript? Can you execute it using `node node_modules/.bin/autorest.cmd`?

anywherepilot: That doesn't work. But the autorest file is also not JavaScript, but bash code. These are the contents of autogen.cmd:

```cmd
@ECHO off
SETLOCAL
CALL :find_dp0

IF EXIST "%dp0%\node.exe" (
  SET "_prog=%dp0%\node.exe"
) ELSE (
  SET "_prog=node"
  SET PATHEXT=%PATHEXT:;.JS;=;%
)

"%_prog%"  "%dp0%\..\autorest\dist\app.js" %*
ENDLOCAL
EXIT /b %errorlevel%
:find_dp0
SET dp0=%~dp0
EXIT /b
```

jp-andre: Okay thanks. Will just call `app.js` instead then.
This new script is really just a convenience anyway, allowing you to run `autorest` only without the re-download of the json file.

PR #833, opened 2020-04-24
Regen client apis

jp-andre:
You have to be super patient to run autorest 2. Autorest 3 still bails out on "Subscription" items from our swagger.
Also `apps` -> `appsOperations` and same for `releases`.
And then fix some invalid formatting and prettier.


src/util/apis/generated/models/index.d.ts
```diff
@@ -47108,8 +47090,7 @@ export interface GetOKResponseAzureSubscription {
   isMicrosoftInternal?: boolean;
 }
 
-export interface
-GetOKResponseModelModelModelModelModelModelModelModelModelModelModelModelModelModel {
+export interface GetOKResponseModelModelModelModelModelModelModelModelModelModelModelModelModelModel {
```
hagould: recognizing that it hasn't been changed, but `GetOKResponseModelModelModelModelModelModelModelModelModelModelModelModelModelModel` is really interesting?

src/commands/apps/create.ts
```diff
@@ -93,7 +93,7 @@ export default class AppCreateCommand extends Command {
     debug(`Creating app with attributes: ${inspect(appAttributes)}`);
     const createAppResponse = await out.progress(
       "Creating app ...",
-      clientRequest<models.AppResponse>((cb) => client.apps.create(appAttributes, cb))
+      clientRequest<models.AppResponse>((cb) => client.appsOperations.create(appAttributes, cb))
```
hagould, line 96: why this change?

anywherepilot: This is strange, looks like a breaking change? One we can handle, of course, as you did.

src/util/apis/generated/models/activeDeviceCountsMonthlyItem.js
```diff
@@ -0,0 +1,59 @@
+/*
```
anywherepilot, line 1: Why so many new files? Did something change in the generator or did these actually appear since we last ran it?

anywherepilot: It's like it added files for all the parameters of all API endpoints.

src/util/apis/generated/models/index.d.ts
jp-andre: I had to remove the newlines. But yeah the names are pretty hilarious.

src/commands/apps/create.ts
jp-andre: I don't know why it changed some of the operations. It could be some new tag in our swagger. I doubt it would be a change from autorest itself since we're using the same major version.

src/util/apis/generated/models/activeDeviceCountsMonthlyItem.js
jp-andre: Dunno why but our swagger json is just much larger and more detailed than it used to be.
It indeed contains the details for all parameters & response types. I can't find what changed that causes this, but it looks fine to me.

We may want to consider shrinking the size of the CLI package by filtering out only the files we need at runtime, but that's my only concern here.

PR #839, opened 2020-04-28
Support app level API tokens

hagould:
These changes are updating the tokens create, tokens list and tokens delete command to account for new arg parameter --type which allows you to choose to create an app API token when an app context or app command parameter is present.
These commands will default to user type tokens unless app is specifically requested.

In order to make and test these changes this PR regenerated the API clients with the app token APIs manually edited to be exposed for generation - in reality they're currently x-internal. 

Once the app API token work is validated and the routes are updated, the API client commit will need to be reverted and regenerated.

`npm run download-swagger` or `npm run autorest`,
then `npm run autorest:gen` 

This PR is dependent on https://github.com/microsoft/appcenter-cli/pull/852
This branch has been merged with master and is ready to go


jp-andre:
Can we separate the regen of API from the code changes? I think we updated the client APIs recently in master branch, maybe you should be able to simply merge master and work from that?

src/commands/tokens/create.ts
```diff
@@ -28,15 +31,37 @@ export default class TokenCreateCommand extends Command {
   @hasArg
   description: string;
 
+  @help("The type of token principal authentication: [" + allPrincipalTypes.join(", ") + "]")
+  @shortName("t")
+  @longName("type")
+  @hasArg
+  @defaultValue("user")
+  public principalType: string;
```
jp-andre, line 39: Any reason to make this `public`?

src/commands/tokens/delete.ts
```diff
@@ -27,22 +33,42 @@ export default class AppDeleteCommand extends Command {
...
+  @help("The type of token principal authentication: [" + allPrincipalTypes.join(", ") + "]")
+  @shortName("t")
+  @longName("type")
+  @hasArg
+  @defaultValue("user")
+  public principalType: string;
+
   async run(client: AppCenterClient): Promise<CommandResult> {
-    const confirmation = await prompt.confirm(`Do you really want to delete the token with ID "${this.id}"`);
+    const tokenLevel = this.principalType === PrincipalType.USER ? principalMessaging.user : principalMessaging.app;
```
jp-andre, line 44: Can we have a validation function that verifies that an acceptable option is passed? We want to have exactly `user` or `app`, or unspecified (default).
Some commands have `validateParameters()`.

src/util/misc/principal-type.ts
```diff
@@ -0,0 +1,11 @@
+export enum PrincipalType {
+  USER = "user",
+  APP = "app",
+}
+
+export const allPrincipalTypes = [PrincipalType.USER, PrincipalType.APP];
+
+export const principalMessaging = {
+  [PrincipalType.USER]: "user type",
+  [PrincipalType.APP]: "app type",
```
jp-andre, line 10: Do we really need this? It seems that simply sticking to PrincipalType value is enough.

test/commands/tokens/create-test.ts
```diff
@@ -0,0 +1,155 @@
+import { expect, use } from "chai";
```
jp-andre, line 1: Thanks so much for creating test cases here! 🎉

test/commands/tokens/create-test.ts
```diff
@@ -0,0 +1,155 @@
...
+      id: fakeId,
+      description: fakeDescription,
+      api_token: "token",
+      scope: "all",
+    });
+
+    const command = new CreateTokenCommand(getCommandArgs([]));
+    const result = await command.execute();
+
+    expect(result.succeeded).to.be.true;
```
jp-andre, line 88: Maybe also check somehow that type is user? Is that possible?

jp-andre:
As our team will be on vacation from Apr 30 to May 5 included, let's wait until full review after May 5 to merge this. Also we'll be expecting a separate PR for the API update.

src/util/misc/principal-type.ts
zacharysierakowski: Agreed, I was going to say the same!

zacharysierakowski:
> Can we separate the regen of API from the code changes? I think we updated the client APIs recently in master branch, maybe you should be able to simply merge master and work from that?

I think we will do this all in one PR actually, if that's okay? We plan to update the PR either on Friday or Monday (May 4th) with the "correct" regen.

src/commands/tokens/create.ts
hagould: This formatting matched other CLI command arguments (I was looking at the distribute ones for reference), but no I didn't actually realize that the previous parameters for these commands weren't marked as public. Would you prefer without `public`?

test/commands/tokens/create-test.ts
hagould: the test only mocks that the user endpoint is being called - which validates that the command defaults to user type token. I might be able to mock the app token endpoint and make sure that it's not being called, but it shouldn't be necessary to cover this case

jp-andre:
Looks like the build failed, let's verify again after the merge conflicts have been resolved.

src/commands/tokens/delete.ts
```diff
@@ -27,22 +33,40 @@ export default class AppDeleteCommand extends Command {
...
+      );
+    } else if (this.principalType === PrincipalType.APP) {
+      const app: DefaultApp = this.app;
+      deleteTokenResponse = await out.progress(
+        tokenMessaging,
+        clientRequest<null>((cb) => client.appApiTokens.deleteMethod(app.ownerName, app.appName, this.id, cb))
+      );
     }
 
+    if (deleteTokenResponse.response.statusCode === 404) {
```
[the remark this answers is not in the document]
hagould: My original intent was to not overly change the behavior of these commands when they are already are being used for user tokens, but I think this shouldn't realistically cause problems

hagould:
the component governance in the build tasks is catching an issue with yarg-parser in the dev gulp dependency. Anyone seen this before or aware of what's best to do here? I don't believe it's something that can be fixed with an audit

Issue #712, opened 2019-11-25
Missing functionality - is posssible to add?


pawel757677:
Hi,
There is lack of two functions:
**1) create app for specific organization**

_example:_
curl --header "Content-Type: application/json" --header "X-API-Token:XXX-XXX" --request POST --data '{"display_name":"DISPLAY_NAME","name":"NAME","os":"Android","platform":"Java"}' https://appcenter.ms/api/v0.1/orgs/{ORGANIZATION}/apps

_solution:_
maybe add option to choose organization for "appcenter apps create" command?

**2) add distribution group**

_example:_
curl --header "Content-Type: application/json" --header "X-API-Token:XXX-XXX" --request POST --data '{"apps":[{"name":"NAME"}]}' https://appcenter.ms/api/v0.1/orgs/{ORGANIZATION}/distribution_groups/{DISTRIBUTION_GROUP}/apps

_solution:_

maybe command "appcenter distribute release -f APK_PATH -g {DISTRIBUTION_GROUP} --app {ORGANIZATION}/{NAME}" should has option to force add distribution group? cause now no working.

jerietve:
Hi @pawel757677, thanks for your request.

Regarding 1, that sounds like a good idea to me. It could be done as you suggest and/or go under `appcenter orgs apps create`.

Regarding 2, the API you call in the example adds an app to a distribution group. The CLI example you write seems to add a distribution group to a release of an app, or to release an app to a distribution group. This is already possible, so perhaps I misunderstand what you want to do. If you want to add a distribution group to an existing release, you can use `distribute releases add-destination`. Please check whether that does what you want.

pawel757677:
Thank you for your response!

Regarding 2. 
You are right, maybe i didn't explain well.
So maybe other example:

Situation: One app, many builds (releases)
So, first i create app (use API or CLI) and now if i run:

`appcenter distribute release -f $APK_PATH -g CustomDistributionGroup --app {ORG}/{NAME}`

Result:

`Error: Could not find group CustomDistributionGroup`

And you are right, i can use commands like this:

```
appcenter distribute release -f $APK_PATH --app {ORG}/{NAME}
LAST_RELEASE_ID=appcenter distribute releases list --app {ORG}/{NAME} | grep ID | head -1 | tr -s ' ' | cut -f2 -d ' '
appcenter distribute releases add-destination -a {ORG}/{NAME} -r $LAST_RELEASE_ID -t group -d CustomDistributionGroup -s
```

[...]

PR #848, opened 2020-05-04
Add `orgs apps create` command.

### LINKED ISSUE #831: Create app in organization
I would like to either:

a) create an app for an organization; or
b) login as an organization 'user' via a token in order to create app in said organization

This doesn't seem possible today:

* `appcenter orgs apps` only offers `list` option
* `appcenter apps create` does not offer option to create on behalf of org

anywherepilot:
It's a clone of `apps create`, except it catches exceptions for nicer user errors and has unit tests. Functional tests probably would have been better, but/since that required creating `orgs delete` and `orgs apps delete` commands as well.

Resolves #831.
Involves #712.

### DIFF
--- a/src/commands/apps/lib/format-app.ts
+++ b/src/commands/apps/lib/format-app.ts
@@ -15,7 +15,7 @@ export function reportApp(app: models.AppResponse): void {
       ["Owner Display Name", "owner.displayName"],
       ["Owner Email", "owner.email"],
       ["Owner Name", "owner.name"],
-      ["Azure Subscription ID", "azureSubscriptionId"],
+      ["Azure Subscription ID", "azureSubscription.subscriptionId"],
     ],
     app
   );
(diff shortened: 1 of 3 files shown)


mrowles:
Looks good 🍕Thanks!

src/commands/apps/lib/format-app.ts
```diff
@@ -15,7 +15,7 @@ export function reportApp(app: models.AppResponse): void {
       ["Owner Display Name", "owner.displayName"],
       ["Owner Email", "owner.email"],
       ["Owner Name", "owner.name"],
-      ["Azure Subscription ID", "azureSubscriptionId"],
+      ["Azure Subscription ID", "azureSubscription.subscriptionId"],
```
yujincat, line 18: Is it a bug fix? I would like to know if this is relevant to the `orgs apps create` command .. (don't think so)

anywherepilot: When the `orgs apps create` command is done creating the app, it writes the app information out to the console using this method, the same way `apps create` does. That method has a bug in it, which I'd like to fix as part of adding the new command.

saikirangit:
Did something messed up with this commit? I dont see this code in master branch??

Issue #842, opened 2020-04-30
Updating notes for existing release


Vacxe:
Does it possible to update release notes for existed release?
`appcenter distribute releases edit` looks very close semantically but responsible for different things.

For now, it possible only with `publish` task.

Will be nice to have a possibility to update existed release with new notes file. Could you please add this functionality?

@jp-andre @anywherepilot

anywherepilot:
It should be on that command, I agree. Looks like we currently only have support for enabling and disabling releases. Editing release notes is possible through the API, so it'll be easy to add.

Vacxe:
@anywherepilot can we add `patch` command for example? Because in the opposite case we should modify `update` command what can actually break back compatibility.

Also how far we can receive this feature? Because it's really important for us. Maybe to push this forward could u send me the links for API?

anywherepilot:
`state` should not be required in my opinion. Actually I'm glad you pointed out this parameter, since I have been dealing with some similar "boolean" type parameters recently and I was thinking it makes sense to have this kind of parameter where you need to write "enabled" or "disabled" explicitly. It's good to know we have a precedent.

You can find our API here: https://openapi.appcenter.ms

The one you'd need to work with is the [PUT operation for updating a release](https://openapi.appcenter.ms/#/distribute/releases_updateDetails).

We'll have a look at your PR. Thanks for taking it upon yourself to add this functionality.

PR #853, opened 2020-05-05
Add update release notes fuctionality

Vacxe:
Close: #842 
Sorry guys but typescript are not my strong side ;) All my best for u

Modification for `appcenter distribute releases edit`

`enabled` is optional now
`release-notes` and `release-notes-file` has been added

Just we require this feature ASAP. Cheers!

### DIFF
--- a/README.md
+++ b/README.md
@@ -84,7 +84,7 @@ Below is the list of commands currently supported by Visual Studio App Center CL
 | `appcenter distribute groups update` | Update existing distribution group |
 | `appcenter distribute releases add-destination` | Distribute an existing release to an additional destination |
 | `appcenter distribute releases delete` | Deletes the release |
-| `appcenter distribute releases edit` | Toggles enabling and disabling the specified release or updating information|
+| `appcenter distribute releases edit` | Toggles enabling and disabling the specified release |
 | `appcenter distribute releases list` | Shows the list of all releases for the application |
 | `appcenter distribute releases show` | Shows full details about release |
 | `appcenter distribute stores list` | Lists all stores of the app |
--- a/README.md
+++ b/README.md
@@ -84,7 +84,7 @@ Below is the list of commands currently supported by Visual Studio App Center CL
 | `appcenter distribute groups update` | Update existing distribution group |
 | `appcenter distribute releases add-destination` | Distribute an existing release to an additional destination |
 | `appcenter distribute releases delete` | Deletes the release |
-| `appcenter distribute releases edit` | Toggles enabling and disabling the specified release |
+| `appcenter distribute releases edit` | Toggles enabling and disabling the specified release or updating information|
 | `appcenter distribute releases list` | Shows the list of all releases for the application |
 | `appcenter distribute releases show` | Shows full details about release |
 | `appcenter distribute stores list` | Lists all stores of the app |
--- a/src/commands/distribute/releases/edit-notes.ts
+++ b/src/commands/distribute/releases/edit-notes.ts
@@ -14,12 +14,12 @@ import { AppCenterClient, clientRequest, models } from "../../../util/apis";
 import { out } from "../../../util/interaction";
 import { inspect } from "util";
 import { handleHttpError } from "../../../util/apis/create-client";
-import * as Pfs from "../../util/misc/promisfied-fs";
+import * as Pfs from "../../../util/misc/promisfied-fs";
 import * as _ from "lodash";
 
 const debug = require("debug")("appcenter-cli:commands:distribute:releases:edit-notes");
 
-@help("Updating relese notes")
+@help("Updating release notes")
 export default class EditReleaseCommand extends AppCommand {
   @help("Release ID")
   @shortName("r")
@@ -70,7 +70,7 @@ export default class EditReleaseCommand extends AppCommand {
       handleHttpError(error, false, "failed to load release details");
     }
 
-    const releaseNotes = await this.getReleaseNotesString()
+    const releaseNotes = await this.getReleaseNotesString();
 
     try {
       debug(`Updating release notes`);
@@ -87,8 +87,7 @@ export default class EditReleaseCommand extends AppCommand {
       debug(`Failed to update the release - ${inspect(error)}`);
       return failure(ErrorCodes.Exception, `failed to update the relese notes`);
     }
-
-    out.text(`Release notes for ${releaseDetails.shortVersion} (${releaseDetails.version}) with id: ${this.releaseId} has been updated`);
+    out.text(`Release ${releaseDetails.shortVersion} (${releaseDetails.version}) with id: ${this.releaseId} has been updated`);
     return success();
   }
 
@@ -113,5 +112,5 @@ export default class EditReleaseCommand extends AppCommand {
     if (!_.isNil(this.releaseNotes) && !_.isNil(this.releaseNotesFile)) {
       throw failure(ErrorCodes.InvalidParameter, "'--release-notes' and '--release-notes-file' switches are mutually exclusive");
     }
-  }  
+  }
 }
--- a/src/commands/distribute/releases/edit-notes.ts
+++ b/src/commands/distribute/releases/edit-notes.ts
@@ -9,17 +9,17 @@ import {
   longName,
   required,
   hasArg,
-  position,
-  name,
 } from "../../../util/commandline";
 import { AppCenterClient, clientRequest, models } from "../../../util/apis";
 import { out } from "../../../util/interaction";
 import { inspect } from "util";
 import { handleHttpError } from "../../../util/apis/create-client";
+import * as Pfs from "../../util/misc/promisfied-fs";
+import * as _ from "lodash";
 
-const debug = require("debug")("appcenter-cli:commands:distribute:releases:delete");
+const debug = require("debug")("appcenter-cli:commands:distribute:releases:edit-notes");
 
-@help("Toggles enabling and disabling the specified release")
+@help("Updating relese notes")
 export default class EditReleaseCommand extends AppCommand {
   @help("Release ID")
   @shortName("r")
@@ -28,26 +28,26 @@ export default class EditReleaseCommand extends AppCommand {
   @hasArg
   public releaseId: string;
 
-  @help("Release state: enabled or disabled")
-  @name("State")
-  @position(0)
-  @required
-  public state: string;
+  @help("Release notes text")
+  @shortName("n")
+  @longName("release-notes")
+  @hasArg
+  public releaseNotes: string;
+
+  @help("Path to release notes file")
+  @shortName("N")
+  @longName("release-notes-file")
+  @hasArg
+  public releaseNotesFile: string;
 
   public async run(client: AppCenterClient): Promise<CommandResult> {
     const app = this.app;
 
+    this.validateParameters();
+
     const releaseId = Number(this.releaseId);
     if (!Number.isSafeInteger(releaseId) || releaseId <= 0) {
       return failure(ErrorCodes.InvalidParameter, `${this.releaseId} is not a valid release id`);
-    }
-
-    const state = this.state;
-    if (state !== "enabled" && state !== "disabled") {
-      return failure(
-        ErrorCodes.InvalidParameter,
-        `"${state}" is not a valid release state. Available states are "enabled" or "disabled".`
-      );
     }
 
     let releaseDetails: models.ReleaseDetailsResponse;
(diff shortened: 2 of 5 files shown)


README.md
```diff
@@ -84,7 +84,7 @@ Below is the list of commands currently supported by Visual Studio App Center CL
 | `appcenter distribute groups update` | Update existing distribution group |
 | `appcenter distribute releases add-destination` | Distribute an existing release to an additional destination |
 | `appcenter distribute releases delete` | Deletes the release |
-| `appcenter distribute releases edit` | Toggles enabling and disabling the specified release |
+| `appcenter distribute releases edit` | Toggles enabling and disabling the specified release or updating information|
```
anywherepilot, line 87: This line is auto generated from the help in the command.

src/commands/distribute/releases/edit.ts
```diff
@@ -29,27 +31,33 @@ export default class EditReleaseCommand extends AppCommand {
   public releaseId: string;
 
   @help("Release state: enabled or disabled")
-  @name("State")
-  @position(0)
-  @required
+  @shortName("s")
```
anywherepilot, line 34: While I agree this should not be a positional parameter (I didn't even know the CLI supported that), this change is breaking. People may have scripts running this operation. Since the command already includes many other named parameters, I doubt it is necessary to make it work? Let's leave this alone. What do you think @jp-andre ?

src/commands/distribute/releases/edit.ts
```diff
@@ -70,23 +78,55 @@ export default class EditReleaseCommand extends AppCommand {
...
+      return this.releaseNotes;
+    }
+  }
+
+  private validateParameters(): void {
+    debug("Checking for invalid parameter combinations");
+    if (!_.isNil(this.releaseNotes) && !_.isNil(this.releaseNotesFile)) {
+      throw failure(ErrorCodes.InvalidParameter, "'--release-notes' and '--release-notes-file' switches are mutually exclusive");
+    }
+    if (!_.isNil(this.state) && this.state === "enabled" || this.state === "disable" ) {
```
anywherepilot, line 128: There's some issues here. `&&` takes precedence over `||`. I don't think this does what you want it to do. The check also needs to be inverted to make sense. And it's checking for `disable` instead of `disabled`. Let's go with this:

`if (!_.isNil(this.state) && this.state !== "enabled" && this.state !== "disabled" )`

src/commands/distribute/releases/edit.ts
```diff
@@ -70,23 +78,55 @@ export default class EditReleaseCommand extends AppCommand {
...
+      throw failure(ErrorCodes.InvalidParameter, "'--state' can be 'enabled' or 'disabled'");
```
anywherepilot, line 129: Consider saying `must be` instead of `can be`.

src/commands/distribute/releases/edit.ts
```diff
@@ -70,23 +78,55 @@ export default class EditReleaseCommand extends AppCommand {
       handleHttpError(error, false, "failed to load release details");
     }
 
+    let enabled: boolean = null
+    if(this.state === "enabled" || this.state === "disabled"){
```
anywherepilot, line 82: Since you validated the parameters, the condition could simply be `!_.isNil(this.state)`.

src/commands/distribute/releases/edit.ts
```diff
@@ -70,23 +78,55 @@ export default class EditReleaseCommand extends AppCommand {
...
+
     try {
-      debug(`Updating release state to "${state}"`);
+      debug(`Updating release`);
       const httpResponse = await out.progress(
-        `${state === "enabled" ? "Enabling" : "Disabling"} the release...`,
+        `Updating release`,
         clientRequest((cb) =>
-          client.releasesOperations.updateDetails(releaseId, app.ownerName, app.appName, { enabled: state === "enabled" }, cb)
+          client.releasesOperations.updateDetails(releaseId, app.ownerName, app.appName, { enabled: enabled , releaseNotes: releaseNotes }, cb)
```
anywherepilot, line 92: Did you test to make sure this doesn't override the release notes with nothing when no the user provides no release notes?

Sadly the CLI cannot distinguish between empty string parameters and the parameter not being provided. We should still allow users to provide an empty file, and use that to set the release notes to nothing, if the API allows that.

src/commands/distribute/releases/edit.ts
```diff
@@ -70,23 +78,55 @@ export default class EditReleaseCommand extends AppCommand {
...
         )
       );
       if (httpResponse.response.statusCode >= 400) {
-        return failure(ErrorCodes.Exception, `failed to ${state === "enabled" ? "enable" : "disable"} the release`);
+        return failure(ErrorCodes.Exception, `failed to update the relese`);
```
anywherepilot, line 96: Typo, same below

test/commands/distribute/releases/edit-test.ts
```diff
@@ -148,7 +141,21 @@ describe("distribute releases edit command", async () => {
...
+      testCommandFailure(executionScope);
+    });
+
+    it("if file and release notes both provided", async () => {
+      // Arrange
+      const expectedErrorMessage = "'--release-notes' and '--release-notes-file' switches are mutually exclusive";
+      const executionScope = _.flow(setupReleaseDetailsResponse, setupUpdateReleaseServiceUnavailableResponse)(Nock(fakeHost));
+
+      // Act
+      const command = new EditReleaseCommand(getCommandArgs([releaseIdOption, fakeReleaseId, "--relese-notes anytext --release-notes-file some/fake/path"]));
```
anywherepilot, line 158: You should investigate why this test passes. The parameter is misspelled: `--relese-notes`. That means it shouldn't throw the expected error.

src/commands/distribute/releases/edit.ts
jp-andre: Correct, we can't have a breaking change here.

src/commands/distribute/releases/edit.ts
jp-andre: More important is that `disabled` is spelled differently in the error and the `if` condition.

anywherepilot: Yeah, see the comment above. The one in the message is the right one.

src/commands/distribute/releases/edit.ts
jp-andre: Note that I don't know if we can have an optional positional parameter, that will need to be investigated.

src/commands/distribute/releases/edit.ts
jp-andre: If the parameter is required then we don't need this check at all.

src/commands/distribute/releases/edit.ts
Vacxe: @jp-andre @anywherepilot that's an issue. Which way is more acceptable: create new command such as ''patch" for example or change current one with breaking back compatibility?

src/commands/distribute/releases/edit.ts
Vacxe: @anywherepilot yes we should. Do u have any clear idea how to implement it?

Vacxe: Or we should to create 2 separate methods such as 'notes update' and 'notes reset'?

src/commands/distribute/releases/edit.ts
Vacxe: Or just 'editnotes' I think that is most clear way which can resolve all the issues with back compatibility and updating to empty notes

jp-andre:
Build fails because of a prettier issue. I think the command to run is `npx prettier --fix`.
(We should have a GitHub Action build for PRs, with no integration tests set up)

src/commands/distribute/releases/edit.ts
anywherepilot: First step is to determine what the API does when it receives empty release notes. It either sets the notes to empty, or ignores it. If it sets the nodes to empty: 1) make sure we don't pass empty just because a user doesn't provide the value, 2) make sure we do pass empty when a user explicitly passes in an empty file. If it ignores the value: no changes necessary, current code is fine.

src/commands/distribute/releases/edit.ts
anywherepilot: I would try to see whether an optional positional parameter works. That will make this command a little weird to work with regarding that one parameter, but at least it still has a discoverable name and is extensible. That's important because there are many more things you can change about a release through the API, that we may want to add to the CLI in the future. The PUT call doesn't have that many options, but the PATCH one does. I don't know why we have that distinction, but to me it'd make sense so call both from this single CLI command.

@jp-andre are you in favor of one edit command that does everything, with one weird positional parameter (if this can be done), or would you rather have separate commands for edit-release-notes and then in the future edit-mandatory and edit-notify-testers, etc.?

Vacxe:
@jp-andre I extracted a separate method for release notes updates

Separate method have these benefits:

- Support backwards compatibility for `edit` command
- If it necessary we can clean a release notes

What do u thinking guys?

src/commands/distribute/releases/edit-notes.ts
```diff
@@ -0,0 +1,117 @@
...
+      return this.releaseNotes;
+    }
+  }
+
+  private validateParameters(): void {
+    debug("Checking for invalid parameter combinations");
+    if (!_.isNil(this.releaseNotes) && !_.isNil(this.releaseNotesFile)) {
+      throw failure(ErrorCodes.InvalidParameter, "'--release-notes' and '--release-notes-file' switches are mutually exclusive");
+    }
+  }
```
jp-andre, line 116: trailing space -- sadly does not show up here in github

test/commands/distribute/releases/edit-notes-test.ts
```diff
@@ -0,0 +1,202 @@
...
+        command = new EditNotesReleaseCommand(getCommandArgs([releaseIdOption, fakeReleaseId, "--release-notes", releaseNotes, "--release-notes-file", releaseFilePath]));
+      } catch (e) {
+        errorMessage = e.message;
+      }
+
+      // Assert
+      expect(command).to.eql(undefined);
+      expect(errorMessage).to.eql(expectedErrorMessage);
+    });  
+
```
jp-andre, line 136: trailing spaces here too

src/commands/distribute/releases/edit-notes.ts
```diff
@@ -0,0 +1,117 @@
...
+        }
+      }
+    } else {
+      return this.releaseNotes;
+    }
+  }
+
+  private validateParameters(): void {
+    debug("Checking for invalid parameter combinations");
+    if (!_.isNil(this.releaseNotes) && !_.isNil(this.releaseNotesFile)) {
```
jp-andre, line 113: What happens when both are unspecified? Note that I don't know the behaviour of the API either.

src/commands/distribute/releases/edit-notes.ts
```diff
@@ -0,0 +1,117 @@
...
+  shortName,
+  longName,
+  required,
+  hasArg,
+} from "../../../util/commandline";
+import { AppCenterClient, clientRequest, models } from "../../../util/apis";
+import { out } from "../../../util/interaction";
+import { inspect } from "util";
+import { handleHttpError } from "../../../util/apis/create-client";
+import * as Pfs from "../../util/misc/promisfied-fs";
```
jp-andre, line 17: `error TS2307: Cannot find module '../../util/misc/promisfied-fs'.`

src/commands/distribute/releases/edit-notes.ts
```diff
@@ -0,0 +1,117 @@
...
+import * as _ from "lodash";
+
+const debug = require("debug")("appcenter-cli:commands:distribute:releases:edit-notes");
+
+@help("Updating relese notes")
```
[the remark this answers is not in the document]
Vacxe: What wrong with me! Quarantine....

src/commands/distribute/releases/edit-notes.ts
Vacxe: If both unspecified It probably will erase the notes. No way to avoid this issue with empty string (or null paramater).

src/commands/distribute/releases/edit-notes.ts
Vacxe: Oh...Type script. Sorry guys but that's my first time with it ;)

jp-andre: That's okay. Just a missing "../" here. Hopefully we can all learn in the process. :)

src/commands/distribute/releases/edit-notes.ts
jp-andre: I think the empty string would be explicit (i.e. `--release-notes ""`), as opposed to `undefined` which would be the result of not passing in either of `--release-notes` or `--release-notes-files`.

jp-andre: And removing the notes by passing in "" is just fine (assuming it's allowed by the API), but what happens when neither option is passed seems unpredictable.

jp-andre:
@Vacxe, if you can run prettier on the code, commit that and fix the remaining unit test, then we may be able to move forward with this PR

test/commands/distribute/releases/edit-notes-test.ts
Vacxe: Btw it's a lint check?

jp-andre: Oh no, just an observation when testing the code locally.

Vacxe:
@jp-andre So build is passed for me locally.  But tests a lot of tests have `Timeout of 2000ms exceeded` error.

jp-andre:
The timeouts can be ignored, but I'm getting this error from your new test.
```
  1 failing

  1) distribute releases edit-notes command
       failed
         if --release-notes and --release-notes-file both specified:
     AssertionError: expected { Object (additionalSupportedOutputFormats, debug, ...) } to deeply equal undefined
      at /Users/jpeg/Code/AC-OpenSource/appcenter-cli/test/commands/distribute/releases/edit-notes-test.ts:136:26
      at Generator.next (<anonymous>)
      at /Users/jpeg/Code/AC-OpenSource/appcenter-cli/test/commands/distribute/releases/edit-notes-test.ts:8:71
      at new Promise (<anonymous>)
      at __awaiter (test/commands/distribute/releases/edit-notes-test.ts:4:12)
      at Context.it (test/commands/distribute/releases/edit-notes-test.ts:119:81)
```

Vacxe:
@jp-andre thanks for your patience for me mate. Now I know a bit more about TS ;) First experience in any case.

Also, could MS guys prepare a minor release after merge when it will be possible (We really need this feature for CI automation)?

jp-andre:
> @jp-andre thanks for your patience for me mate. Now I know a bit more about TS ;) First experience in any case.

No problem.

> Also, could MS guys prepare a minor release after merge when it will be possible (We really need this feature for CI automation)?

I will have to see about it with the rest of my team but we understand your situation.

src/commands/distribute/releases/edit-notes.ts
jp-andre: @Vacxe let's make sure that at least one of the two parameters is specified. I tested the command line and specifying no param does not do anything, while specifying `-n ''` resets the notes to nothing (displayed as -).

src/commands/distribute/releases/edit-notes.ts
```diff
@@ -0,0 +1,116 @@
...
+import { AppCenterClient, clientRequest, models } from "../../../util/apis";
+import { out } from "../../../util/interaction";
+import { inspect } from "util";
+import { handleHttpError } from "../../../util/apis/create-client";
+import * as Pfs from "../../../util/misc/promisfied-fs";
+import * as _ from "lodash";
+
+const debug = require("debug")("appcenter-cli:commands:distribute:releases:edit-notes");
+
+@help("Updating release notes")
```
anywherepilot, line 22: Note to self: fix tense.

src/commands/distribute/releases/edit-notes.ts
```diff
@@ -0,0 +1,116 @@
...
+    try {
+      debug(`Updating release notes`);
+      const httpResponse = await out.progress(
+        `Updating release notes`,
+        clientRequest((cb) =>
+          client.releasesOperations.updateDetails(releaseId, app.ownerName, app.appName, { releaseNotes: releaseNotes }, cb)
+        )
+      );
+      if (httpResponse.response.statusCode >= 400) {
+        return failure(ErrorCodes.Exception, `failed to update the relese notes`);
```
anywherepilot, line 84: Note to self: fix typo, same below

Vacxe: Oh...Sorry

PR #857, opened 2020-05-07
Fix copy paste debug typo

Vacxe:
@anywherepilot @jp-andre


anywherepilot:
I checked the codebase: this is the only broken one.

PR #895, opened 2020-06-01
Fix "isOrg" check in test command

zacharysierakowski:
Get app info to determine if owner is org instead of listing all orgs. This will allow App API token usage with the `test` commands.

### DIFF
--- a/src/commands/test/lib/test-cloud-uploader.ts
+++ b/src/commands/test/lib/test-cloud-uploader.ts
@@ -72,13 +72,8 @@ export class TestCloudUploader {
   }
 
   public async uploadAndStart(): Promise<StartedTestRun> {
-    const orgs = await getOrgsNamesList(this._client);
-    let isOrg = false;
-    for (const org of orgs) {
-      if (org.name === this._userName) {
-        isOrg = true;
-      }
-    }
+    const app = await this._client.appsOperations.get(this._userName, this._appName);
+    const isOrg = app.owner.type === "org";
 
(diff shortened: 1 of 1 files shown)


anywherepilot:
This unrelated test is failing. https://github.com/microsoft/appcenter-cli/blob/master/test/functional/commands/distribute/Release.Tests.ps1

Maybe there's an issue in the integration environment at the moment.

zacharysierakowski:
> This unrelated test is failing. https://github.com/microsoft/appcenter-cli/blob/master/test/functional/commands/distribute/Release.Tests.ps1
> 
> Maybe there's an issue in the integration environment at the moment.

Yea that definitely looks unrelated. Let me know if it happens again next time you run it!

zacharysierakowski:
I figured out how to run those failing tests locally and they are failing on `master` as well :thinking_face:
image

anywherepilot:
I was looking into it yesterday. If I run those commands by myself one by one on my own account, it works fine. I then grabbed the API key for the test account for the key vault, as the pipeline does, and ran the steps against it as well. That also worked fine. I don't know what's going on. If you need to get this in today, feel free to disable the test. The test by itself is not that important; I'll enable it again afterward after figuring out what the issue is.

zacharysierakowski:
Thanks for helping out @anywherepilot. It doesn't necessarily need to be merged today. I just didn't want to lose track of it especially if there was something needed from my end. If you are okay with disabling the test to get the merge across that'd be great.

jp-andre:
Sorry for the delay. As soon as #903 is merged we can move forward with this.

PR #905, opened 2020-06-10
Add build version and build number to request

annakocheshkova:
We forgot to implement a body stub during feature implementation. Adding the required params.


jp-andre:
Please improve the PR description. Why is this required? Also, can we have a test?

annakocheshkova:
@jp-andre we already have a tests for that, thanks to this test the bug was found. This is just a forgotten stub during feature implementation, let me add that to the description.

PR #910, opened 2020-06-10
Extract a filename from path

vvechkanov:
When uploading a file with a path, the full path was sent to the server as a filename. Fix this, by extracting the only filename for sending to the server.

### DIFF
--- a/src/commands/distribute/lib/mc-fus-uploader/mc-fus-uploader.ts
+++ b/src/commands/distribute/lib/mc-fus-uploader/mc-fus-uploader.ts
@@ -29,6 +29,10 @@ export class McFile implements McFusFile {
     return stats["size"];
   }
 
+  get fileName(): string {
+    return this.name.replace(/^.*[\\\/]/, '');
+  }
+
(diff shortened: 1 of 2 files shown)


src/commands/distribute/lib/mc-fus-uploader/mc-fus-uploader.ts
```diff
@@ -29,6 +29,10 @@ export class McFile implements McFusFile {
     return stats["size"];
   }
 
+  get fileName(): string {
+    return this.name.replace(/^.*[\\\/]/, '');
```
MatkovIvan, line 33: @vvechkanov @russelarms Can we just replace `name` implementation?

vvechkanov: Yes, I think we can. Add a getter for a name, and use property inside a class. If js support this (I assume it is), we could do it.

MatkovIvan, line 33: path.basename?

PR #913, opened 2020-06-11
Fix package-lock.json file.

vvechkanov:
Fix package-lock.json file, that was changed during feature development.


Jamminroot:
Thanks for fixing that

PR #920, opened 2020-06-11
Fix all the unit tests

annakocheshkova:
This PR modifies all the unit tests with regards to the feature.

PR #921, opened 2020-06-11
Use token parameter in release command

annakocheshkova:
This PR makes sure that a `--token` parameter is used when specified.
This also fixes several unit tests which were failing if user is not logged in on the machine.


annakocheshkova:
The CI fails because of lint which is fixed by another PR #920 . I've run this on another branch with all the fixes and all the the tests pass.

PR #922, opened 2020-06-12
How to run functional tests

anywherepilot:


contributing.md
```diff
@@ -224,8 +224,9 @@ you tests to keep things consistent across the entire codebase.
 #### test/functional
 
 Tests that call the appcenter-cli from PowerShell to execute against an App Center internal test environment live here.
-Files called \*Tests.ps1 run automatically. The [Pester test runner](https://github.com/pester/Pester/wiki/Pester) is used
-to run the tests. You can run theses tests locally by running the Main.ps1 file in test/functional using PowerShell.
+Files called \*Tests.ps1 run automatically. The [Pester test runner](https://github.com/pester/Pester/wiki/Pester) runs the tests.
+
+You can run theses tests locally, keeping in mind they will log on to your account and perform destructive operations. Consider using a dedicated testing account. Run the tests by [creating an API token for App Center](https://docs.microsoft.com/appcenter/api-docs/#creating-an-app-center-app-api-token) and using it to call Invoke-FunctionalTests.ps1.
```
jp-andre, line 229: An easy way to create a token is to use the CLI :D `appcenter login`.

jp-andre: Also this link points to App Token, which I don't think will work here since we create the app.

jp-andre: It turns out the CLI tokens created interactively with cli (appcenter login) can't be used with `--token`. You need a full access token from the portal for `--token`. But the comment about the link still stands valid.

anywherepilot: Thanks for catching that. I should have noticed the page was updated since we introduced the new app type tokens. It links to the user one now.

contributing.md
```diff
@@ -224,8 +224,9 @@ you tests to keep things consistent across the entire codebase.
 #### test/functional
 
 Tests that call the appcenter-cli from PowerShell to execute against an App Center internal test environment live here.
```
yujincat, line 226: So I guess only people have access to INT will be able to run the test?

anywherepilot: Wow. You are totally right. I cannot believe I didn't think of that. Thank you! That makes them useless for external people who want to test their work. I'll revisit this PR and add an option to the code to run it against production.

PR #923, opened 2020-06-12
Fix functional tests

annakocheshkova:
The release retrieval was failing because the file is dummy and cannot be extracted on server side.


annakocheshkova:
This CI also fails because of lint which is fixed by another PR #920 . I've run this on another branch with all the fixes and all the the tests pass.

src/commands/distribute/lib/mc-fus-uploader/mc-fus-uploader.ts
```diff
@@ -16,7 +16,7 @@ import {
 import * as fs from "fs";
 import fetch from "node-fetch";
 import { MimeTypes } from "./mc-fus-mime-types";
-const Path = require('path');
+const Path = require("path");
```
MatkovIvan, line 19: We need to change it to `import` (not in this PR)

PR #909, opened 2020-06-10
Use App Center File Uploader Service for release uploading

russelarms:
- Adapt mc-fus-uploader library for node environment
- Change some of the release uploading methods in accordance with the web version

TODO list:
- [X] Fix lint warnings
- [x] Fix ts config and related warnings
- [X] Add and rework tests for mc-fus-uploader library
- [X] Fix other tests
- [x] Fix CI build
- [x] Fix filename [bug](https://msmobilecenter.visualstudio.com/Mobile-Center/_workitems/edit/81099)

[AB#80602](https://msmobilecenter.visualstudio.com/Mobile-Center/_boards/board/t/IvanM-Team/Backlog%20Items/?workitem=80602)

PR #929, opened 2020-06-15
Address fus-library feedback

russelarms:
Address feedback from https://github.com/microsoft/appcenter-cli/pull/909

### DIFF
--- a/src/commands/distribute/lib/mc-fus-uploader/mc-fus-uploader-types.ts
+++ b/src/commands/distribute/lib/mc-fus-uploader/mc-fus-uploader-types.ts
@@ -1,6 +1,6 @@
 export interface McFusUploader {
-  Start(file: McFusFile): void;
-  Cancel(): void;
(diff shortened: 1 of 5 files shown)


src/commands/distribute/lib/mc-fus-uploader/mc-fus-uploader-types.ts
```diff
@@ -1,6 +1,6 @@
 export interface McFusUploader {
-  Start(file: McFusFile): void;
-  Cancel(): void;
+  start(file: McFusFile): void;
+  cancel(): void;
 }
 
 export enum McFusMessageLevel {
```
jp-andre, line 6: Not a big deal but field names also start with lowercase, usually. This looks like C# :)

russelarms: Fixed :)

PR #931, opened 2020-06-16
CI: Add steps to verify test result files exist

jp-andre:


azure-pipelines.yml
```diff
@@ -39,7 +39,7 @@ jobs:
     condition: and(succeeded(), ne(variables['Build.Reason'], 'PullRequest'))
 
   - script: |
-      npm run test:ci || true
+      npm run test:ci || [ -f './test-result.xml' ]
```
anywherepilot, line 42: What does this do? The script below implies that `[ -f <path> ]` returns whether that path exists. So here we either run the tests or check that the results exist? To continue in case of test failure and publish the failed results, for better viewing?

yujincat: I would like to be explained more on this :-) What is the difference between `test-result.xml` and `testresult.xml` and `test-results.xml` (with 's')? The `test-results.xml` file is found in task `PublishTestResults@2` which is right after like 42.

PR #952, opened 2020-06-22
Fix search path for functional test results

jp-andre:


jp-andre:
Looks like I missed an `s` in the other file.

PR #962, opened 2020-06-26
loadReleaseIdUntilSuccess retries on 500 error response

AnastasiaKubova:
Add a check for the status code of the response and process it accordingly to avoid retrying when receiving error response.

### DIFF
--- a/src/commands/distribute/release.ts
+++ b/src/commands/distribute/release.ts
@@ -469,7 +469,7 @@ export default class ReleaseBinaryCommand extends AppCommand {
           "x-api-token": accessToken,
         },
       });
-      if (response.status < 200 || response.status > 300) {
+      if (response.status < 200 || response.status >= 300) {
         throw failure(ErrorCodes.Exception, `failed to get release id with HTTP status: ${response.status} - ${response.statusText}`);
       }
       return await response.json();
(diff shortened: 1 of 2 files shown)


src/commands/distribute/release.ts
```diff
@@ -463,6 +469,9 @@ export default class ReleaseBinaryCommand extends AppCommand {
           "x-api-token": accessToken,
         },
       });
+      if (response.status < 200 || response.status > 300) {
```
Jamminroot, line 472:
```suggestion
      if (response.status < 200 || response.status >= 300) {
```

test/commands/distribute/release/release-test.ts
```diff
@@ -454,6 +454,37 @@ describe("release command", () => {
...
+            )
+          )
+        )
+      );
+      skippedRequestsScope = setupSuccessfulGetUploadResponse(
+        setupSuccessfulCreateReleaseResponse(setupSuccessfulAddGroupResponse(Nock(fakeHost)))
+      );
+    });
+
+    it("should fail during get release when HTTP status isn't 200", async () => {
```
Jamminroot, line 475:
```suggestion
    it("should fail during get release when HTTP status isn't 2xx", async () => {
```

AnastasiaKubova:
Yes

PR #978, opened 2020-07-07
Use ac-fus-lib as npm package

annakocheshkova:
It was decided to treat mc-fus as a separate library and publish it to npm without creating a separate repo. 
This PR makes it a proper library and moves to the root of the repo (for visibility).
Also did some minor renaming to differentiate it with the original mc-fus library for web.

This PR also adds an additional step to CI that builds the library.
[AB#81826](https://msmobilecenter.visualstudio.com/Mobile-Center/_workitems/edit/81826)


jp-andre:
> It was decided to treat mc-fus as a separate library and publish it to npm without creating a separate repo.

I don't feel very comfortable with this approach. Why not create a separate open source repo?
If this lib can be used in another client, it also shouldn't be in the CLI. Could it be shared by the portal? Otherwise, why have a separate npm lib?

appcenter-file-upload-client-node/azure-pipelines.yml
```diff
@@ -0,0 +1,23 @@
...
+  inputs:
+    versionSpec: '10.x'
+  displayName: 'Install Node.js'
+
+- script: |
+    npm install
+    npm install -g eslint
+    npm run build
+    npm run test
+  displayName: 'Build and test'
```
jp-andre, line 23: What is this file for? Are we creating a new pipeline in azure? I don't think this approach can work.

package.json
```diff
@@ -44,7 +45,8 @@
   "files": [
     "package-lock.json",
     "bin",
-    "dist"
+    "dist",
+    "appcenter-file-upload-client-node/out"
```
jp-andre, line 49: I think this is missing various files, like package.json and package-lock.json.

package.json
```diff
@@ -17,7 +17,8 @@
   },
   "preferGlobal": true,
   "scripts": {
-    "build": "npm run lint && gulp build:raw",
+    "build-ac-fus": "cd appcenter-file-upload-client-node && npm i && npm run build",
```
jp-andre, line 20: `npm i` should happen during install, not during build

appcenter-file-upload-client-node/.eslintignore
```diff
@@ -0,0 +1,8 @@
+node_modules
+
+# compiled files
+out
+build
+
+# db migration files
+db
```
jp-andre, line 8: newline please

appcenter-file-upload-client-node/.prettierignore
```diff
@@ -0,0 +1,10 @@
+node_modules
+
+# compiled files
+out
+build
+
+# db migration files
+db
+
+random-folder
```
jp-andre, line 10: newline

package-lock.json
```diff
@@ -758,6 +758,1969 @@
...
       }
     },
+    "appcenter-file-upload-client-node": {
+      "version": "file:appcenter-file-upload-client-node",
+      "requires": {
+        "abort-controller": "^3.0.0",
+        "node-fetch": "^2.6.0"
+      },
+      "dependencies": {
+        "@appcenter/eslint-config-appcenter": {
```
jp-andre, line 768: this is not acceptable: we can't use a private package from an open source repo

annakocheshkova:
@jp-andre Resolved the feedback, following the test instructions now works for me even if I install the `.tgz` package globally.

jp-andre:
That's awesome! Thanks so much for taking the time. We will have another look on Monday (tmr is our day of learning).

anywherepilot:
Works indeed.

Issue #953, opened 2020-06-22
Flag to disable parallel uploads introduced in version 2.6.0


aaratn:
We are using travis-ci and suddenly our distribute command stopped working until we found out that there is a new concurrent upload feature introduced in version 2.6.0. To fix this issue we downgraded the version to previous one i.e 2.5.4.

We are looking for an option where we can disable con-currency for distribution release.

andysousa:
same here, i could not get an upload to work for the life of me behind a corporate proxy with the 2.6.0 release, had to go back to 2.5.4.  I am not sure what changed... but its either not honoring the proxy correctly or something else.

Jamminroot:
Hey, @aaratn, @andysousa, thanks for getting in touch. 
I've tried to repro the problem with my own repo (travis-CI with private repo), and it did not quite reproduce, and release was correctly created. Is there anything else I can be missing? Have a look at configs I made to test (apk uploaded as binary):
```yml
language: node_js
node_js:
  - node
script:
  - bash scripts/deploy.sh
```
```bash
#!/bin/bash
npm install -g appcenter-cli
appcenter login --token $TOKEN --debug true
appcenter distribute release -f test.apk -r "Travis-CI Distribution via appcenter-cli" --app $APP_NAME --group Public --debug true
```
Also, can you run appcenter-cli with debug logging turned on? Just append `--debug true` to the command. Don't forget to omit sensitive information (secrets, tokens, etc).
Sorry for such a delay in response.

aaratn:
@Jamminroot 

Can you try doing same for ios? I was testing this for ios builds

andysousa:
this happens behind a corporate proxy for me... i will get you the debug output in a little bit

Jamminroot:
@andysousa, sure, waiting for those.
@aaratn did test that - also no problems.
@andysousa, @aaratn Need more context for further investigation. So far it looks more like a problem with the related changes we made recently... Can you guys provide me the exact command (with args) you are running? And again - don't forget to edit sensitive info out before posting here.

Jamminroot:
@andysousa @aaratn can you check that specifying token explicitly as a command line argument rather than env var helps?
[This PR](https://github.com/microsoft/appcenter-cli/pull/979) fixes recent bug with getting token from env - that would explain why travis-ci worked for me.

andysousa:
[...]

PR #979, opened 2020-07-09
Fix release upload with token as env var

### LINKED ISSUE #973: Not able to create new release when access token is given using environment variable
Version 2.6.1

I am trying to publish a new version of app to AppCenter, and the distrubute command fails with error:

> Error: failed to create release upload for ....ipa. Missing valid authentication token. Correlation ID: 06e24f3a-a11d-44ca-b100-4c75978d56ef

The same command works with CLI version 2.5.4 (verified by downgrading)

Auth token was provided using environment variable in both cases.

jp-andre:
Closes #973


Jamminroot:
Might be related to #953

PR #996, opened 2020-07-29
Add a flag for settng the project name when running "release-react"

matt-oakes:
This is needed when you want to make use of the logic to automatically detect the current iOS version, but the `package.json`'s `name` does not match the name of the iOS project path. This is common in monorepo setups where it might be called something like `mobile-app`.

This PR adds a new argument for the command allowing you to manually set the project name. It still uses all of the same logic for finding the version.

I have also removed the `projectRoot` argument for the `getReactNativeProjectAppVersion` as it was unused and was always `undefined` whenever the function was called.


DmitriyKirakosyan:
Hi @matt-oakes this issue was addressed here: #1497.
I'm closing this PR.

PR #997, opened 2020-07-30
Enable Dependabot on appcenter-file-upload-client-node

anywherepilot:


.github/dependabot.yml
```diff
@@ -1,9 +1,14 @@
...
     # Look for `package.json` and `lock` files in the `root` directory
     directory: "/"
     # Check the npm registry for updates every weekday
     schedule:
       interval: "daily"
+  # Enable version updates for npm on the inner appcenter-file-upload-client-node package
+  - package-ecosystem: "npm"
+    directory: "/appcenter-file-upload-client-node"
+    schedule:
+      interval: "daily"
```
jp-andre, line 14: maybe we should switch to weekly :)

anywherepilot: I am getting a bit tired of getting updates right after I'm done merging stuff :D. I'll see if I can set it to run on Wednesdays only, both of them.

anywherepilot: You can't choose the day on a per-repo basis, just set what "weekly" means to your organization. Still yeah let's just set them both to weekly to get a bit less notifications and less spread in the channel.

PR #1004, opened 2020-07-30
Fixes passing token/env into release task from group publish

jphenow:
Fixes in issue where `distribute release` with `--token` works but `distribute groups publish` does not (with a 403) due to not forwarding params from the groups/publish task.


jp-andre:
Oh nice catch -- I guess we have the same issue with stores publish?

jphenow:
Possibly! We can fix as part of this PR. I'll glance at that command as well

jphenow:
Oh I didn't get the update on the page until I commented lol. Thanks for doing that stores change @anywherepilot!

PR #1017, opened 2020-08-10
Fixes Error: spawn node_modules/hermesvm/osx-bin/hermes ENOENT when running codepush release with 0.63 react native

a7ul:
Renaming hermes to hermesc.

Reference issue: https://github.com/microsoft/react-native-code-push/issues/1886


anywherepilot:
@a7ul we'll take #1018 instead, since it takes into account which version you're running with. We're checking whether the name should change for Windows as well, as in this PR.

PR #1018, opened 2020-08-11
[CodePush] Fix Hermes path for new version

andreidubov:
**Issue:**  Hermes path, it changed in the RN 0.63 release (because they upgraded Hermes from 0.4.0 to 0.5.0).

**Solution:** Update path for new versions.

**Related issue:** #1017 [#1886](https://github.com/microsoft/react-native-code-push/issues/1886)


anywherepilot:
We should update the Windows command too: https://github.com/facebook/hermes/commit/284ae5e70d6bb2c7d98e859e74907c68ae6abae0

Issue #1007, opened 2020-07-31
EACCES: permission denied during npm install 


xeivieni:
Hi, 

I'm having an issue installing appcenter-cli@2.6.4 using npm 6.14.6. 

Here is the trace of the error : 
```
appcenter-cli@2.6.4 preinstall /usr/local/lib/node_modules/appcenter-cli
> cd appcenter-file-upload-client-node && npm i

npm ERR! code EACCES
npm ERR! syscall scandir
npm ERR! path /root/.npm/_logs
npm ERR! errno -13
npm ERR!
npm ERR! Your cache folder contains root-owned files, due to a bug in
npm ERR! previous versions of npm which has since been addressed.
npm ERR!
npm ERR! To permanently fix this problem, please run:
npm ERR!   sudo chown -R 65534:0 "/root/.npm"
glob error [Error: EACCES: permission denied, scandir '/root/.npm/_logs'] {
  errno: -13,
  code: 'EACCES',
  syscall: 'scandir',
  path: '/root/.npm/_logs'
}
npm ERR! code EACCES
npm ERR! syscall open
npm ERR! path /root/.npm/_cacache/index-v5/ce/ae/fd93e677a614f57b325ca5f22842c3e00f2071dd70d44f24a4b413d6fe52
npm ERR! errno -13
npm ERR!
npm ERR! Your cache folder contains root-owned files, due to a bug in
npm ERR! previous versions of npm which has since been addressed.
npm ERR!
npm ERR! To permanently fix this problem, please run:
npm ERR!   sudo chown -R 65534:0 "/root/.npm"
npm WARN webpack-cli@3.3.12 requires a peer of webpack@4.x.x but none is installed. You must install peer dependencies yourself.

npm ERR! code ELIFECYCLE
npm ERR! errno 243
npm ERR! appcenter-cli@2.6.4 preinstall: `cd appcenter-file-upload-client-node && npm i`
npm ERR! Exit status 243
npm ERR!
npm ERR! Failed at the appcenter-cli@2.6.4 preinstall script.
npm ERR! This is probably not a problem with npm. There is likely additional logging output above.

npm ERR! A complete log of this run can be found in:
npm ERR!     /root/.npm/_logs/2020-07-31T09_44_16_062Z-debug.log
```

I actually have no error when installing 2.6.3

Thanks

bertucci13:
Same for me. The new version seems corrupted.

okalman:
+1

[...]

PR #1021, opened 2020-08-14
Consume appcenter-file-upload-client-node as a package

russelarms:
The [original PR's]( https://github.com/microsoft/appcenter-cli/pull/978) description:

> It was decided to treat mc-fus as a separate library and publish it to npm without creating a separate repo.
> This PR makes it a proper library and moves to the root of the repo (for visibility).
> Also did some minor renaming to differentiate it with the original mc-fus library for web.
> This PR also adds an additional step to CI that builds the library.

The changes to the original PR:
- [x] fixed vulnerabilities in the ac-fus-lib dev dependencies
- [x] ac-fus-lib connected as an npm dependency in the main package.json, it should fix the [linux issue](https://github.com/microsoft/appcenter-cli/issues/1007)
- [x] ac-fus-lib types are imported from the npm dependency

Tested:
- [x] manually
- [x] with `npm test` and `npm test:ci`
- [x] ci tests

Related work items:
[AB#82467](https://msmobilecenter.visualstudio.com/Mobile-Center/_boards/board/t/IvanM-Team/Backlog%20Items/?workitem=82467)
[AB#82466](https://msmobilecenter.visualstudio.com/Mobile-Center/_boards/board/t/IvanM-Team/Backlog%20Items/?workitem=82466)


src/commands/distribute/lib/mc-fus-uploader/readme.md
```diff
@@ -1,8 +0,0 @@
-This directory contains a NodeJS port of the library used to upload releases to App Center.
-
-The library leverages App Center File Upload Service and allows uploading many chunks (slices of a release binary) in parallel. It takes the binary file, slices it in 4mb chunks, and uploads them simultaneously.
```
MatkovIvan, line 3: Should we keep some of this desc in the new readme?

package.json
```diff
@@ -44,7 +46,10 @@
   "files": [
     "package-lock.json",
     "bin",
-    "dist"
+    "dist",
+    "appcenter-file-upload-client-node/out",
+    "appcenter-file-upload-client-node/package.json",
+    "appcenter-file-upload-client-node/package-lock.json"
```
MatkovIvan, line 52: This is wrong. These files must not be directly included in CLI npm package.

jp-andre:
Thanks for working on this. There is still a build error with integration tests (the ones in powershell).

annakocheshkova:
@jp-andre it has not appeared on the 2nd run. Probably a flaky one, since we're talking to the real AppCenter API.

azure-pipelines.yml
```diff
@@ -40,16 +40,36 @@ jobs:
...
+    workingDirectory: appcenter-file-upload-client-node
+
+  - task: PublishTestResults@2
+    inputs:
+      testResultsFormat: 'JUnit'
+      testResultsFiles: 'test-results.xml'
+      searchFolder: '$(Build.SourcesDirectory)/appcenter-file-upload-client-node'
+      failTaskOnFailedTests: true
+      testRunTitle: 'appcenter-file-upload-client-node unit tests'
+    displayName: 'Verify appcenter-file-upload-client-node unit test result'
```
jp-andre, line 72: Could we have an "install test" here? Where we validate that `npm install -g` works on the package? Ideally we should be verifying the scenario that didn't work (linux with sudo and no nvm), but I'm not sure this is easily feasible within this pipeline.

package.json
```diff
@@ -17,7 +17,8 @@
   },
   "preferGlobal": true,
   "scripts": {
-    "build": "npm run lint && gulp build:raw",
+    "build-ac-fus": "cd appcenter-file-upload-client-node && npm run build",
+    "build": "npm run build-ac-fus && npm run lint && gulp build:raw",
```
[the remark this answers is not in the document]
annakocheshkova: I don't think we need it. For CI, we have [this step](https://github.com/microsoft/appcenter-cli/pull/1021/files#diff-fec826feae04e51c0d94076385408bdcR55), similar to building the cli itself.

package.json
```diff
@@ -103,6 +104,7 @@
   },
   "dependencies": {
     "abort-controller": "^3.0.0",
+    "appcenter-file-upload-client-node": "1.0.0",
```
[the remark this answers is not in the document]
annakocheshkova: I don't think they are somehow related.

azure-pipelines.yml
annakocheshkova: We have npm install [here](https://github.com/microsoft/appcenter-cli/pull/1021/files#diff-fec826feae04e51c0d94076385408bdcR55), and if it doesn't work, then testing step will fail. I don't see similar verification steps for the `npm i` for CLI so not sure why we need it?

jp-andre: Sorry, I meant installing the appcenter-cli package itself onto the system. i.e. `npm pack && npm install -g appcenter-cli*.tgz`.
What was failing in the previous attempt was installing this package on a user's computer.

package.json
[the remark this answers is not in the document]
jp-andre: Okay. Too bad we did not coordinate those efforts better, but it's OK :)

appcenter-file-upload-client-node/package.json
```diff
@@ -0,0 +1,45 @@
...
+  "description": "File Uploader Client for Node",
+  "homepage": "https://github.com/microsoft/appcenter-cli",
+  "author": "Microsoft",
+  "license": "MIT",
+  "repository": {
+    "type": "git",
+    "url": "https://github.com/microsoft/appcenter-cli"
+  },
+  "scripts": {
+    "build": "npm run lint && npm run build:tsc",
```
jp-andre, line 13: I didn't mean to remove build-ac-fus entirely, only to remove it from build. But that's ok. :)

anywherepilot (approved):
This overall looks great to me. I would have put the .gitignore stuff in a separate file in the sub folder but the effect is the same. JP and I were thinking of adding a package installation test just as a followup to the incident that caused the revert of the previous version of this change. It's not related to the new package, but it would be a nice hardening followup to what was caused by a previous release of this change.

Dependabot is still enabled on the sub folder, and we'll handle those updates going forward. We'll also release new versions of the package as part of that process.

jp-andre:
> This overall looks great to me. I would have put the .gitignore stuff in a separate file in the sub folder but the effect is the same. JP and I were thinking of adding a package installation test just as a followup to the incident that caused the revert of the previous version of this change. It's not related to the new package, but it would be a nice hardening followup to what was caused by a previous release of this change.

It's included in this PR as I requested it, though this is only a simple test here, that wouldn't have caught the linux-only sudo issue.

jp-andre (approved):

MatkovIvan (approved):

PR #1034, opened 2020-08-28
Make appcenter-cli debugging easier

wisdeom:
**Updates**
1. `npm run build` runs the linter every time wasting time when debugging. So, created `npm run compile` to use it when debugging.
2. `node2` is outdated, so updated it to `node`.  
3. Add a section to explain how to use debugging configurations defined in `launch.json` in the contribuding.md. 
4. Created a new debugging config, `Launch without building sourcemaps` to use when there's no change in source code. It reduces time to build sourcemaps as it's not necessary.

### DIFF
--- a/contributing.md
+++ b/contributing.md
@@ -117,6 +117,7 @@ There are a bunch of scripts in package.json file. Here's what they are and what
 | `npm run clean`      | Cleans up any compilation output                                                       |
 | `npm run autorest`   | Regenerate the HTTP client libraries. Downloads required tools as part of this process |
 | `npm run lint`       | Runs the linter over the codebase                                                      |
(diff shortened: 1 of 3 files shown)


contributing.md
```diff
@@ -137,6 +138,15 @@ The gulpfile.js file contains the following targets that can be called manually
 | `copy-generated-client` |              | Copies the generated HTTP client code to dist                           |
 | `prepublish`            | `prepublish` | Runs the `clean` and `build:raw` tasks before publishing to npm         |
 
+## Debugger
+
+Debugging configurations are stored in launch.json file. Here're what they are and what they do:
```
anywherepilot, line 143: It's probably obvious once you're in that file, but consider adding an example for how to debug a specific command, showing how to add its arguments.

anywherepilot (approved):

PR #1052, opened 2020-09-10
appcenter-file-upload-client-node version to 1.1.1

anywherepilot:
Dependency updates only

### DIFF
--- a/appcenter-file-upload-client-node/package.json
+++ b/appcenter-file-upload-client-node/package.json
@@ -1,6 +1,6 @@
 {
   "name": "appcenter-file-upload-client-node",
-  "version": "1.1.0",
+  "version": "1.1.1",
   "description": "File Uploader Client for Node",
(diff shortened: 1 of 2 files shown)


jp-andre (approved):

appcenter-file-upload-client-node/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "appcenter-file-upload-client-node",
-  "version": "1.1.0",
+  "version": "1.1.1",
```
ajaykn, line 3: is this some bug fix ?

anywherepilot: Dependency updates. We only change the patch version for that.

jp-andre: Dependency updates only

ajaykn: ah cool

PR #1069, opened 2020-09-24
Version to 2.7.1

anywherepilot:
Dependency updates and actually uptaking the thing that should have been in 2.7.0

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,6 +1,6 @@
 {
   "name": "appcenter-cli",
-  "version": "2.7.0",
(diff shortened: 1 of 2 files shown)


jp-andre:
Why not wait for acfus 1.1.2 before making this release?

jp-andre (approved):

anywherepilot:
Because I don't want to manually put it in when it's just dependency updates. I don't mind saving some dev time as a trade-off to being two more weeks behind.

PR #1104, opened 2020-10-15
[code-push] [release-react] fix version with caret and react-native custom build

n-kurdiukov-akvelon:
This fix resolves these issues:
https://github.com/microsoft/react-native-code-push/issues/1676
https://github.com/microsoft/react-native-code-push/issues/1939

I've used `semver.coerce()` to clear version string. Now `mobiletechvn/react-native#v0.63.2.fix-shadow-node` or `^0.63.2` will transform into `0.63.2`. And Hermes version will be checked correctly.

### DIFF
--- a/src/commands/codepush/lib/validation-utils.ts
+++ b/src/commands/codepush/lib/validation-utils.ts
@@ -21,7 +21,7 @@ export function isValidRollout(rollout: number): boolean {
 }
 
 export function isLowVersion(v1: string, v2: string): boolean {
-  return semver.compare(v1, v2) === -1 ? true : false;
+  return semver.compare(semver.coerce(v1).version, v2) === -1;
 }
 
 export async function isValidDeployment(client: AppCenterClient, app: DefaultApp, deploymentName: string): Promise<boolean> {
--- a/src/commands/codepush/lib/validation-utils.ts
+++ b/src/commands/codepush/lib/validation-utils.ts
@@ -1,3 +1,4 @@
+import * as semver from "semver";
 import { AppCenterClient, models, clientRequest } from "../../../util/apis";
 import { DefaultApp } from "../../../util/profile";
 
--- a/src/commands/codepush/lib/validation-utils.ts
+++ b/src/commands/codepush/lib/validation-utils.ts
@@ -1,4 +1,3 @@
-import * as semver from "semver";
 import { AppCenterClient, models, clientRequest } from "../../../util/apis";
 import { DefaultApp } from "../../../util/profile";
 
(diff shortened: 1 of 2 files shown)


src/commands/codepush/lib/validation-utils.ts
```diff
@@ -21,7 +21,7 @@ export function isValidRollout(rollout: number): boolean {
 }
 
 export function isLowVersion(v1: string, v2: string): boolean {
-  return semver.compare(v1, v2) === -1 ? true : false;
+  return semver.compare(semver.coerce(v1).version, v2) === -1;
```
snowpardx, line 24: Thanks for removing ternary operator this really makes code more clean. 

By the way doing `semver.coerce` here is probably not the thing we want, because it transforms the single argument, but not another one which may lead to some strange results if not the fixed version provided as second argument.

My assumption here is that we should have either have  both arguments transformed or even better not doing transformation here at all and instead transform result of `getReactNativeVersion()` here
https://github.com/microsoft/appcenter-cli/blob/118367c5303034acc4028b942e4134f68db767fa/src/commands/codepush/lib/react-native-utils.ts#L442

n-kurdiukov-akvelon: I did it because `validation-utils.ts` already has the imported module "semver" and `isLowVersion()` is used only in `const hermesExecutableName = ` line.  If we will move this fix into `react-native-utils.ts`, the better way, I think, is to delete the `isLowVersion()` function and change `react-native-utils.ts` 442 line into `const hermesExecutableName = semver.compare(semver.coerce(getReactNativeVersion()).version, "0.63.0") === -1 ? "hermes" : "hermesc";` or move `isLowVersion` in `react-native-utils.ts` and change it into isLowThan63version (with the function logic changing)

n-kurdiukov-akvelon: or change `isLowVersion()` body into `return semver.compare(semver.coerce(v1).version, semver.coerce(v2)) === -1;` but it will be an unnecessary premature optimization.

n-kurdiukov-akvelon: @snowpardx , what is the best option you think?

ruslan-bikkinin (dismissed):

src/commands/codepush/lib/validation-utils.ts
n-kurdiukov-akvelon: removed isLowVersion function

ruslan-bikkinin (changes requested):
@n-kurdiukov-akvelon could you please write some tests to cover problematic cases?

src/commands/codepush/lib/validation-utils.ts
snowpardx: I'm totally fine with the current revision, where the helper function which have good context only around call to RN version check is inlined. Thank you!

snowpardx (approved):

andreidubov (approved):

Krasavinigor (approved):

anywherepilot (approved):
Looks good to me. Since this is such a common function built into semver, I don't see a need for testing edge cases here.

Issue #911, opened 2020-06-10
React Native bundling fails but CLI still proceeds to release


RishabhKarnad:
When running the `codepush release-react` subcommand, if there is a failure when creating the React Native bundle due to a lack of heap space in the node process, a release is still created.

#### Steps to reproduce:
1. Use `CODE_PUSH_NODE_ARGS` to reduce node max_old_space_size to a small enough number, so that the bundler runs out of memory. e.g. `export CODE_PUSH_NODE_ARGS='--max_old_space_size=256'`
2. Run `appcenter codepush release-react`. e.g. `appcenter codepush release-react -a <appName> -d <deployment>`

#### Expected result
Since the bundler fails, a release should not be created

#### Actual result
A release is created. This causes CI builds to pass. The size of the bundle is only a few bytes and results in a crash and rollback when installed in the RN app.

#### react-native version
0.62

#### react-native-code-push version
6.2.0

#### Example logs
```
+ appcenter codepush release-react -a [REDACTED] --token [REDACTED] --quiet --sourcemap-output --output-dir ./build-ios --deployment-name Staging
Detecting ios app version:
Using the target binary version value "1.0.0" from [REDACTED].
Running "react-native bundle" command:
node node_modules/.bin/react-native bundle --assets-dest build-ios/CodePush --bundle-output build-ios/CodePush/main.jsbundle --dev false --entry-file index.js --platform ios --sourcemap-output build-ios/CodePush/main.jsbundle.map
Welcome to React Native!
                Learn once, write anywhere

<--- Last few GCs --->
[3170:0x102862000]   616509 ms: Mark-sweep 1321.7 (1444.6) -> 1306.3 (1445.1) MB, 3073.1 / 0.0 ms  (average mu = 0.122, current mu = 0.042) allocation failure scavenge might not succeed
[3170:0x102862000]   620305 ms: Mark-sweep 1321.4 (1445.6) -> 1306.1 (1445.1) MB, 3609.9 / 0.0 ms  (average mu = 0.085, current mu = 0.049) allocation failure scavenge might not succeed
<--- JS stacktrace --->
==== JS stack trace =========================================
    0: ExitFrame [pc: 0x30de0465be3d]
Security context: 0x0e3069c9e6c1 <JSObject>
[...]

PR #1115, opened 2020-10-22
[CodePush] Cancel release for failed bundle

Krasavinigor:
**Issue**: appcenter-cli creates release when React Native bundling is failed
**Related issue**: #911 
**Solution**: Reject release process when React Native bundling is failed

### DIFF
--- a/src/commands/codepush/lib/react-native-utils.ts
+++ b/src/commands/codepush/lib/react-native-utils.ts
@@ -321,7 +321,6 @@ export function runReactNativeBundleCommand(
     });
 
     reactNativeBundleProcess.stderr.on("data", (data: Buffer) => {
-      console.error(data.toString().trim());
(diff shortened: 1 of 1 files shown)


RishabhKarnad (commented):
Hey @Krasavinigor, thanks for the PR. I think this might work, but my concern is that if the buffer doesn't contain the entire error message, the `stderr.on` may be called multiple times.

Maybe it should be changed to `stderr.once`?

Even then I see a possibility of the entire error message not being displayed, and parts of the stack trace may get left out.

There might also be non-fatal errors that get logged to `stderr`, which could now end up causing builds to fail.

It's weird that the RN bundler process does not close with code 1, if the bundler has an error like the one I faced.

Krasavinigor:
I'm going to close PR because it doesn`t fully fix this error. For now, please feel free to reopen it if you have any questions.

PR #1192, opened 2020-12-15
[Distribution] Do not call abortReleaseUpload API when uploading file failed

dhei:
Context: We are deprecating `PATCH /v0.1/apps/{owner_name}/{app_name}/release_uploads/{upload_id}` API ([openapi.appcenter.ms swagger link](https://openapi.appcenter.ms/#/distribute/releaseUploads_complete)). When calling App Center CLI with incorrect build number or incorrect build version (see command below), the CLI will call the deprecated PATCH API with `aborted` on file upload failure. Note this bug only reproduces when both build version and build number parameters are provided but at least one of them is invalid.

`appcenter distribute release --file <localFilePath> --app <appName> --group <groupName> --build-version <build-version> --build-number <build-number>`

We investigated the distribute backend and discovered the `aborted` release upload status is never used in any meaningful way. We should stop using `aborted` status.

Solution: On the client-side, CLI will no longer call any aborted API on upload failure. The server side will be fixed in a different PR.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "appcenter-cli",
-  "version": "2.7.4",
+  "version": "2.7.3",
   "description": "Command line tool for Visual Studio App Center",
   "keywords": [
     "microsoft",
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "appcenter-cli",
-  "version": "2.7.3",
+  "version": "2.7.4",
   "description": "Command line tool for Visual Studio App Center",
   "keywords": [
     "microsoft",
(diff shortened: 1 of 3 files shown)


dhei:
FYI @atkhalsa

atkhalsa (approved):

andrii-z4i:
@dhei could you please share the PR which actually fixes the backend behavior
>Solution: On the client-side, CLI will no longer call any aborted API on upload failure. The server side will be fixed in a different PR.

I think server-side should go first in such case and only if it has no issues we can complete this PR.

andrii-z4i (approved):

MatkovIvan (approved):

PR #1215, opened 2021-01-15
[CodePush] Add limit for request

Krasavinigor:
**Issue**: `deployment list` command gets the list of the deployments (which works flawlessly as this is one request) and then for each deployment tries to get metrics info if `latest_release` field of the deployment is not empty by sending request-per-deployment simultaneously. The server in that case returns 429 error "Too many requests".

**Solution**: Limit the number of requests per time.


andreidubov (approved):

alexandergoncharov (approved):

russelarms (approved):

christophermark:
Thank you! ❤️ 
I just tested and deployment lists work well now for large amounts.

PR #1317, opened 2021-03-10
CodePush React: Adds a flag to  specify a custom React Native project name

danielrhodes:
### Discussion
When creating a React Native project, one can specify a name which becomes the iOS project name. By default this name is saved into `package.json`, and this CLI tool will attempt to read from there to derive the name.

However, it is not uncommon for the `package.json` name to change (e.g. to fit into yarn workspaces).

Therefore, it seems appropriate that a developer would want to change this name and override the automatically discovered name when necessary.

### Solution

Add a flag to the CLI which lets a developer specify the name used when creating a React Native project.

### Example Usage
```sh
> appcenter codepush release-react -a foo/bar --deployment-name Staging -m --description "Foo Bar" --project-name app
```

#### Expected Result:

If the React Native name was `app` when the project was created, the XCode Project will be named `app.xcodeproj`. Thus specifying `--project-name app` will point the CLI to `ios/app.xcodeproj` when building to auto-detect the version.

### DIFF
--- a/src/commands/codepush/lib/react-native-utils.ts
+++ b/src/commands/codepush/lib/react-native-utils.ts
@@ -19,6 +19,7 @@ export interface VersionSearchParams {
   plistFilePrefix: string;
   gradleFile: string;
   buildConfigurationName: string;
+  projectName: string;
 }
 
 interface XCBuildConfiguration {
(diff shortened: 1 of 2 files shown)


jgillick:
+1

DmitriyKirakosyan:
Hi guys. This was fixed here: https://github.com/microsoft/appcenter-cli/pull/1497.
I'm closing this PR.

PR #1397, opened 2021-04-27
Add verification before loadActiveUsersStatistics

Jamminroot:
This PR adds verification before performing loadActiveUsersStatistics AB#86189

### DIFF
--- a/src/commands/analytics/audience.ts
+++ b/src/commands/analytics/audience.ts
@@ -108,7 +108,7 @@ export default class AudienceCommand extends AppCommand {
     if (this.activeUsers && appBuild) {
       promises.push(this.loadActiveUsersStatistics(statistics, client, app, startDate, endDate, appVersion, appBuild));
     } else if (this.activeUsers){
-      debug('Please provide app version to get active users statistics.');
+      out.text('Please provide app version to get active users statistics.');
     }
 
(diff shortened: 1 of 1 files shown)


src/commands/analytics/audience.ts
```diff
@@ -105,8 +105,10 @@ export default class AudienceCommand extends AppCommand {
       promises.push(this.loadLanguagesStatistics(statistics, client, app, startDate, endDate, appVersion));
     }
 
-    if (this.activeUsers) {
+    if (this.activeUsers && appBuild) {
       promises.push(this.loadActiveUsersStatistics(statistics, client, app, startDate, endDate, appVersion, appBuild));
+    } else if (this.activeUsers) {
+      out.text("Please provide app version to get active users statistics.");
```
DmitriyKirakosyan, line 111: Shouldn't it be a warning instead?

Jamminroot: `out` only supports `text()` for messages and `help()` for help output.
Other options are: `debug()` (not a part of `out`), which would only show up if debug is enabled (meaning that most customers won't be able to see the message), and `console.log`\`console.warn`, which ignore `--quiet` flags. 
So, `out.text()` is the closest to what we need.

DmitriyKirakosyan (approved):

PR #1398, opened 2021-04-28
Return previous value for PROMISE_CONCURRENCY

Krasavinigor:
We should return PROMISE_CONCURRENCY to prev version because it's not an effective strategy for the customer.

[AB#86388](https://msmobilecenter.visualstudio.com/Mobile-Center/_workitems/edit/86388)
[AB#85283](https://msmobilecenter.visualstudio.com/Mobile-Center/_workitems/edit/85283)

### DIFF
--- a/src/commands/codepush/deployment/list.ts
+++ b/src/commands/codepush/deployment/list.ts
@@ -18,7 +18,7 @@ import { promiseMap } from "../../../util/misc/promise-map";
 import { formatDate } from "./lib/date-helper";
 
 const debug = require("debug")("appcenter-cli:commands:codepush:deployments:list");
-const PROMISE_CONCURRENCY = 20;
+const PROMISE_CONCURRENCY = 30;
 
 @help("List the deployments associated with an app")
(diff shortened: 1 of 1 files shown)


andreidubov (approved):

alexandergoncharov (approved):

DmitriyKirakosyan:
@Krasavinigor I noticed your message "Unfortunately, sometimes customer still has this issue" in the original issue. So does this change fix the issue completely or does it reduce its probability of happening?

Krasavinigor:
@DmitriyKirakosyan, thank you for the review! Both shots miss. These changes #1388 contains the fix for two issues: 
- adding the new field metrics for the undefined property `latestRelease`
- [AB#85283](https://msmobilecenter.visualstudio.com/Mobile-Center/_workitems/edit/85283)

The customer informed us that the second issue isn't resolved, so I want to return the value to the previous state.

DmitriyKirakosyan (approved):

MatkovIvan (approved):

PR #1756, opened 2021-12-07
Fix Hermes enabled check for iOS

bryansum:
Noticed that despite Hermes being enabled I wasn't seeing the `appcenter-cli` try to create Hermes bytecode for the build -- this was because the check for Hermes was too specific -- it expects:
```
:hermes_enables => true
```
when equally valid is new Ruby syntax:
```
hermes_enabled: true
```

This change adds the latter case as well to the test.

### DIFF
--- a/src/commands/codepush/lib/react-native-utils.ts
+++ b/src/commands/codepush/lib/react-native-utils.ts
@@ -498,7 +498,7 @@ export function getiOSHermesEnabled(podFile: string): boolean {
 
   try {
     const podFileContents = fs.readFileSync(podPath).toString();
-    return /:hermes_enabled(\s+|\n+)?=>(\s+|\n+)?true/.test(podFileContents);
+    return /(:hermes_enabled(\s+|\n+)?=>(\s+|\n+)?true|hermes_enabled:(\s+|\n+)?true)/.test(podFileContents);
   } catch (error) {
     throw error;
   }


DmitriyKirakosyan:
Hi @bryansum , thank you for your contribution!

Could you please link the source where this `:hermes_enables => true` -> `hermes_enabled: true` replacement was introduced? I'm still seeing the old syntax in react-native docs.

bryansum:
It hasn't been replaced anywhere, it's just that that syntax is also valid Ruby, so it would seem worthwhile covering this case as well. For instance my `Podfile` looks like:
```
  use_react_native!(
    path: config[:reactNativePath],
    # to enable hermes on iOS, change `false` to `true` and then install pods
    hermes_enabled: true
  )
```
and indeed has Hermes enabled. 

Generally speaking, this method of detection seems generally a bit brittle in any case -- what might make more sense would be to instead read the `Podfile.lock` and look for the `hermes-engine` string, since that's what the `hermes_enabled` flag enables anyway: https://github.com/facebook/react-native/blob/main/scripts/react_native_pods.rb#L85 I can make that change too if that makes more sense.

AnatolyPristensky (approved):

DmitriyKirakosyan:
Hi @bryansum , since this is a utility method and it can be used anywhere, I would like to avoid relying on a generated file which may not exist in project at some point.
Let's keep this changes at this moment, and if you still feel this check should be done in lock file, please create a separate PR or issue and we will have a deeper look into int.
Thank you again for the contribution!

Issue #1820, opened 2022-02-01
Cannot patch latest release when only setting --disabled false


jdoltar:
### **Description**
We prefer to release with `--disabled` initially, then enable it at a later time (patch with `--disabled false`). We are unable to patch the latest release in this way due to the following error:
```
Error: At least one property must be specified to patch a release.
```

It looks like this is a bug with [this check](https://github.com/microsoft/appcenter-cli/blob/b6619a90bcc048eac95843c1d1fa6819577c1d9b/src/commands/codepush/patch.ts#L80-L82):
```typescript
if (!this.targetBinaryRange && !this.isDisabled && !this.isMandatory && !this.description && !this.rollout) {
  return failure(ErrorCodes.Exception, "At least one property must be specified to patch a release.");
}
```

`!this.isDisabled` resolves to `true` as if it was not provided.

It seems the type of `isDisabled` should be nullable, then the above can instead check that it's not null, since both `true` and `false` are valid values.

### **Repro Steps**
Please list the steps used to reproduce your issue.
1. `appcenter codepush release-react -a SomeOwner/MyApp-iOS --disabled `
2.  `appcenter codepush patch -a SomeOwner/MyApp-iOS --disabled false`

### **Expected behavior**
The latest codepush release should be enabled (downloadable by end users) 

### **Details**
1. Is there a particular beacon/command that you are having trouble with?
   appcenter codepush patch
2. Which version of the App Center CLI are you using?
    2.10.8
3. Which OS version did you experience the issue on?
    macOS BigSur 11.6
4. What command line/shell are you using?
   zsh
5. What version of Node.js and NPM/Yarn are you using?
    node v14.15.4
    npm v6.14.10
    yarn v1.22.10
6. Additionally, you can provide verbose logs of a CLI command by using `DEBUG="appcenter-cli:*"` before the command, for example, `DEBUG="appcenter-cli:*" appcenter apps list`. Include the logs here, don't forget to **omit** any personal/sensitive data.

<details>
  <summary>Verbose log output</summary>

```
yarn run v1.22.10
$ /Users/jondoltar/p/blvd/Mobile/node_modules/.bin/appcenter codepush patch --app SomeOwner/MyApp-iOS Staging --disabled false --token <REDACTED>
2022-02-01T02:34:59.185Z appcenter-cli:util:commandline:option-decorators @help decorator called with 3 arguments: [
  { [Symbol(OptionParameters)]: { debug: [Object] } },
  'debug',
[...]

PR #1823, opened 2022-02-04
Update condition for isDisabled and isMandatory properties

aleksandr-dorofeev:
Fix Cannot patch latest release when only setting --disabled false issue 

Related issue: [1820](https://github.com/microsoft/appcenter-cli/issues/1820/)

### DIFF
--- a/src/commands/codepush/patch.ts
+++ b/src/commands/codepush/patch.ts
@@ -12,6 +12,7 @@ import {
   hasArg,
   position,
   name,
+  defaultValue,
 } from "../../util/commandline";
 import { out } from "../../util/interaction";
 import { inspect } from "util";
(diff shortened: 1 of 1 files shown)


src/commands/codepush/patch.ts
```diff
@@ -77,10 +81,15 @@ export default class PatchCommand extends AppCommand {
     const app = this.app;
     let release: models.CodePushRelease;
 
-    if (!this.targetBinaryRange && !this.isDisabled && !this.isMandatory && !this.description && !this.rollout) {
+    if (!this.targetBinaryRange && _.isNil(this.isDisabled) && _.isNil(this.isMandatory) && !this.description && !this.rollout) {
       return failure(ErrorCodes.Exception, "At least one property must be specified to patch a release.");
     }
 
+    if (_.isNil(this.isDisabled) || _.isNil(this.isMandatory)) {
```
DmitriyKirakosyan, line 88: this condition is not correct

DmitriyKirakosyan:
LGTM. This PR also fixes a bug when a codepush patch changed release to `enabled` state if it was disabled before and patched with no `--disabled` flag specified. With this fix the release keeps the old value for `enabled` and `mandatory` flags if they were no specified 👍

DmitriyKirakosyan (approved):

PR #2004, opened 2022-07-12
Codepush: Handle subprocess error cases where exitCode is null.

hpello:
This PR aims at fixing issue #911.

In subprocesses' `'close'` event handlers:

```javascript
.on("close", (exitCode: number, signal: string) => { ... }
```

By checking that `exitCode` is `!== 0`, we handle the cases where `exitCode` is `null` and `signal` is defined.
References:
- https://nodejs.org/api/child_process.html#class-childprocess
- https://stackoverflow.com/questions/39159312/child-process-exits-with-code-null

PR #2017, opened 2022-07-21
Codepush: Handle subprocess error cases where exitCode is null.

DmitriyKirakosyan:
This PR aims at fixing issue #911.

In subprocesses' `'close'` event handlers:

```javascript
.on("close", (exitCode: number, signal: string) => { ... }
```

By checking that `exitCode` is `!== 0`, we handle the cases where `exitCode` is `null` and `signal` is defined.
References:
- https://nodejs.org/api/child_process.html#class-childprocess
- https://stackoverflow.com/questions/39159312/child-process-exits-with-code-null

Reopened from https://github.com/microsoft/appcenter-cli/pull/2004 with fixes in ui tests.

### DIFF
--- a/src/commands/codepush/lib/electron-utils.ts
+++ b/src/commands/codepush/lib/electron-utils.ts
@@ -76,9 +76,9 @@ export function runWebPackBundleCommand(
       console.error(data.toString().trim());
     });
 
-    webpackProcess.on("close", (exitCode: number) => {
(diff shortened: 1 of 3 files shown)


dorofeeevs (approved):

aleksandr-dorofeev (approved):

hpello:
🎉 😊 👏

PR #2099, opened 2022-09-27
[release-react] Do not generate sourcemaps if only `output-dir` is specified

### LINKED ISSUE #1451: Exclude *.map files from package to reduce size (5mb => 1.4mb)
**Is your feature request related to a problem? Please describe.**

I use the next command to bandle package and deploy it to AppCenter

```sh
npx appcenter codepush release-react --app "my.user/MyApp-Android" \
   --deployment-name "test" \
   --sourcemap-output \ # https://docs.microsoft.com/en-us/appcenter/distribution/codepush/cli#source-map-output-parameter
   --output-dir "./build-android"
```

I use `--sourcemap-output` to upload "*.map" files to my error tracking service (Sentry: https://docs.sentry.io/platforms/react-native/manual-setup/codepush/)

But a package is increased 3 times ( and this new size is compressed .map file) 

**Describe the solution you'd like**

prevent add debug files into package from the box

**Describe alternatives you've considered**

**Additional context**

hduprat:
[release-react] Do not generate sourcemaps if only `output-dir` is specified and neither `sourcemap-output` nor `sourcemap-output-dir` are specified.

Please tell me if you have any remarks or suggestions about this PR.

### DIFF
--- a/src/commands/codepush/release-react.ts
+++ b/src/commands/codepush/release-react.ts
@@ -189,9 +189,8 @@ export default class CodePushReleaseReactCommand extends CodePushReleaseCommandB
       out.text('\n"sourcemap-output-dir" argument will be ignored as "sourcemap-output" argument is provided.\n');
     }
 
-    if ((this.outputDir || this.sourcemapOutputDir) && !this.sourcemapOutput) {
-      const sourcemapDir = this.sourcemapOutputDir || this.updateContentsPath;
-      this.sourcemapOutput = path.join(sourcemapDir, this.bundleName + ".map");
+    if (this.sourcemapOutputDir && !this.sourcemapOutput) {
+      this.sourcemapOutput = path.join(this.sourcemapOutputDir, this.bundleName + ".map");
     }
 
     this.targetBinaryVersion = this.specifiedTargetBinaryVersion;


DmitriyKirakosyan:
@hduprat Thanks a lot for the contribution. We will test the change and merge if it works as expected.

DmitriyKirakosyan:
@hduprat I've checked your PR. This will end up creating a source map in a wrong location, and thus it will not be included in the codepush release.

I think we should avoid creating source map at all if we don't need it.
Currently we create a source map if source-dir was provided, which might be incorrect actually. However I'm worried about breaking existing behaviour. If we change it, it may break many set-ups which rely on source map being uploaded within the codepush release.
On the other hand we are not creating source map if we don't specify `output-dir` param, which looks inconsistent.
Also, according to the parameter description `Path to where the sourcemap for the resulting bundle should be written. If omitted, a sourcemap will not be generated`, we should not create a source map if it was not provided.

So, let's remove `outputDir` from the `if` condition and prevent creating a source map if `sourcemap-output` and `sourcemap-output-dir` were not specified.

MikhailSuendukov:
Hi @hduprat! Just a friendly ping.

hduprat:
Hi @DmitriyKirakosyan ,

> Currently we create a source map if source-dir was provided, which might be incorrect actually. However I'm worried about breaking existing behaviour. If we change it, it may break many set-ups which rely on source map being uploaded within the codepush release.

Do you know if setups like this actually exist? The issue #1451 is precisely that sourcemaps are uploaded with the release, and thus downloaded on the apps (and it takes ~70% of bundle size). I can't think of a valid use of sourcemaps inside an app so if you have some examples in mind, it would help me a lot. Let's discuss this in the issue itself!

> So, let's remove `outputDir` from the `if` condition and prevent creating a source map if `sourcemap-output` and `sourcemap-output-dir` were not specified.

Ok, that means the command itself has a bug (generating sourcemaps if `outputDir` is specified, even if `sourcemap-output` and `sourcemap-output-dir` are omitted. I will focus on fixing this bug for now.

DmitriyKirakosyan (approved):

DmitriyKirakosyan (approved):

Priskorn (approved):

Issue #2112, opened 2022-10-11
codepush release-react doesn't enable Hermes in some cases


youedd:
### **Description**
`appcenter codepush release-react` command only checks if `buildGradle["project.ext.react"]` contains the line `enableHermes: true` to enable hermes [(source code)](https://github.com/microsoft/appcenter-cli/blob/ecc2bf98eb193ac0f1bc348080e58d6130445e5d/src/commands/codepush/lib/react-native-utils.ts#L489) , which leads to scenarios where hermes is not enabled when it should be.

Same problem for iOS.

### **Repro Steps**
Please list the steps used to reproduce your issue.
1. `npx react-native init test`
2. in android/app/build.gradle change ` enableHermes: true` to `enableHermes: "cond" == "cond"` 
3. appcenter codepush release-react ...
4. the output bundle is javascript instead of byte code

### **Expected behavior**
The output bundle should be byte code

### **Details**
1. Is there a particular beacon/command that you are having trouble with?
    - `appcenter codepush release-react`
2. Which version of the App Center CLI are you using?
    - 2.11.0
3. Which OS version did you experience the issue on?
    - macOS Monterey 12.2.1
4. What command line/shell are you using?
    - zsh
5. What version of Node.js and NPM/Yarn are you using?
    - node v16.15.1, npm 8.11.0

aleksandr-dorofeev:
Hi @youedd , thank you for reaching out and for the contribution!
After we test the changes, I will update this issue.

AnatolyPristensky:
@youedd, thank you for your contribution!
Your PR has been tested and merged. Closing this issue.

PR #2113, opened 2022-10-11
Add --use-hermes arg to bypass hermes checks and force enable it

youedd:
With this PR users can bypass hermes checks and force enable it using `--use-hermes` arg

Fixes https://github.com/microsoft/appcenter-cli/issues/2112

### DIFF
--- a/src/commands/codepush/release-react.ts
+++ b/src/commands/codepush/release-react.ts
@@ -125,7 +125,7 @@ export default class CodePushReleaseReactCommand extends CodePushReleaseCommandB
 
   @help("Enable hermes and bypass automatic checks")
   @longName("use-hermes")
-  public useHermes: string | string[];
+  public useHermes: boolean;
 
   private os: string;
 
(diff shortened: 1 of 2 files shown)


youedd:

aleksandr-dorofeev (dismissed):

src/commands/codepush/release-react.ts
```diff
@@ -123,6 +123,10 @@ export default class CodePushReleaseReactCommand extends CodePushReleaseCommandB
   @hasArg
   public extraHermesFlags: string | string[];
 
+  @help("Enable hermes and bypass automatic checks")
+  @longName("use-hermes")
+  public useHermes: string | string[];
```
DmitriyKirakosyan, line 128: Why it is not a `boolean`?

youedd: Hi @DmitriyKirakosyan, thanks for the review!
You're right it should be a `boolean`

dorofeeevs (approved):

PR #2172, opened 2022-12-19
fix(rnutils): fixed hermes path

sjors98:
Hi!

I've stumbled upon an issue where the hermesCommand wouldn't work with the `appcenter codepush release-react` command.
Our project is setup in a monorepo which causes this line to fail https://github.com/microsoft/appcenter-cli/blob/6c4e476a24280146750c1eea75bb5e933cf6c127/src/commands/codepush/lib/react-native-utils.ts#L545

And the way our build.gradle is setup also caused errors in https://github.com/microsoft/appcenter-cli/blob/6c4e476a24280146750c1eea75bb5e933cf6c127/src/commands/codepush/lib/react-native-utils.ts#L550
Since we used gradle variables there which aren't supported in the react-native-utils since they don't get parsed properly

Anyway, I fixed the first issue with a simple require and path.dirname so it always links to the proper react-native path even though there is a monorepo setup or something alike.

### DIFF
--- a/src/commands/codepush/lib/react-native-utils.ts
+++ b/src/commands/codepush/lib/react-native-utils.ts
@@ -541,7 +541,8 @@ async function getHermesCommand(gradleFile: string): Promise<string> {
     }
   };
   // Hermes is bundled with react-native since 0.69
-  const bundledHermesEngine = path.join("node_modules", "react-native", "sdks", "hermesc", getHermesOSBin(), getHermesOSExe());
+  const reactNativePath = path.dirname(require.resolve('react-native'));
+  const bundledHermesEngine = path.join(reactNativePath, 'sdks', 'hermesc', getHermesOSBin(), getHermesOSExe());
   if (fileExists(bundledHermesEngine)) {
     return bundledHermesEngine;
   }


sjors98:
For those that are interested I have made a patch with patch-package for now but would like to see this fixed in the package itself ofcourse: 
```
diff --git a/node_modules/appcenter-cli/dist/commands/codepush/lib/react-native-utils.js b/node_modules/appcenter-cli/dist/commands/codepush/lib/react-native-utils.js
index c2e62f6..b06ce13 100644
--- a/node_modules/appcenter-cli/dist/commands/codepush/lib/react-native-utils.js
+++ b/node_modules/appcenter-cli/dist/commands/codepush/lib/react-native-utils.js
@@ -452,7 +452,8 @@ function getHermesCommand(gradleFile) {
             }
         };
         // Hermes is bundled with react-native since 0.69
-        const bundledHermesEngine = path.join("node_modules", "react-native", "sdks", "hermesc", getHermesOSBin(), getHermesOSExe());
+        const reactNativePath = path.dirname(require.resolve('react-native'));
+        const bundledHermesEngine = path.join(reactNativePath, 'sdks', 'hermesc', getHermesOSBin(), getHermesOSExe());
         if (fileExists(bundledHermesEngine)) {
             return bundledHermesEngine;
         }
```

DmitriyKirakosyan:
Hi @sjors98 , thank you for the contribution!

I tested your changes, and there are issues I faced:

1. It will crash, if you use `appcenter` as a global command (without `npx`)
2. It will crash, if it fails to find the `react-native` module
3. It will display whole path, with username etc, which might be an exposure of sensitive information.
4. Lint :)

So, in general I would like this solution be rather a fallback if the current code fails to find the module. Also I would like it to be try-catched properly. As for the exposing of the full path, I'm not sure. Ideas?

DmitriyKirakosyan:
Hi @sjors98 , friendly ping. Would you like to continue working on your PR?

sjors98:
I'm sorry @DmitriyKirakosyan I totally forgot this PR. I don't think I'll have the time to work on this PR anymore. We switched form yarn to pnpm which resolved our monorepo issues for us and patches like these aren't necessary anymore.
If anyone finds this and thinks this is a solution for them, it sort of is but refactoring your monorepo setup to pnpm will probably work better ;p

PR #2268, opened 2023-04-05
Update rollout error message

abhiin1947:
Update rollout error message

Looks like rollout is supposed to be a value between 1 & 100 but the error message is a little misleading. Fix the message and the grammer.


abhiin1947:

AnatolyPristensky (approved):

AnatolyPristensky:
@abhiin1947, thank you for your contribution! I've tested it and can confirm that changes are valid. 
Merging.

PR #2284, opened 2023-04-15
Use node to resolve react-native package path

youedd:
We were facing issues with the appcenter cli after migrating our react-native project to a mono-repository.

The cli is looking for react-native package is in the project node_modules folder instead of the root node_modules, which leads to errors.

This PR aims to solve that by using node's `require.resolve` to find the package path.

### DIFF
--- a/src/commands/codepush/lib/react-native-utils.ts
+++ b/src/commands/codepush/lib/react-native-utils.ts
@@ -12,6 +12,7 @@ const plist = require("plist");
 const g2js = require("gradle-to-js/lib/parser");
 const properties = require("properties");
 import * as childProcess from "child_process";
+import { directoryExistsSync } from "../../../util/misc/fs-helper";
 
 export interface VersionSearchParams {
   os: string; // ios or android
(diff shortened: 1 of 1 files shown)


DmitriyKirakosyan:
Hi @youedd , thank you for the contribution! If I understand correctly, this solution will work only if appcenter-cli is added to your react-native project as a dependency. What if we use it as a globally installed tool?

DmitriyKirakosyan:
@youedd  thank you for the clarification. You are right.

However, the fallback doesn't work. If the process throws an error, it will return empty string in `result.stdout` and `path.dirname("")` returns `"."`. Then it will confirm that the path `"."` exists and will return it. I think that in addition to checking if the path exists, we also need to check that `result.status` is `0`.

Also, probably worth adding a timeout to `spawnSync` call, in case it freezes for some reason.

AnatolyPristensky:
Merged with a needful changes as https://github.com/microsoft/appcenter-cli/pull/2305

PR #2339, opened 2023-05-30
Update react-native-utils.ts

### LINKED ISSUE #2212: Hermes not working with react native 0.71 on Android
Command `appcenter codepush ...` working but not show message
> Converting JS bundle to byte code via Hermes ...

- appcenter-cli: 2.31.1
- react-native: 0.71
- react-native-code-push: 7.1.0 with patch [https://github.com/microsoft/react-native-code-push/issues/2418#issuecomment-1426484628](https://github.com/microsoft/react-native-code-push/issues/2418#issuecomment-1426484628)

BANG88:
Fixes #2212

### DIFF
--- a/src/commands/codepush/lib/react-native-utils.ts
+++ b/src/commands/codepush/lib/react-native-utils.ts
@@ -499,7 +499,6 @@ async function parseGradlePropertiesFile(gradleFile: string) {
 }
 
 export async function getAndroidHermesEnabled(gradleFile: string): Promise<boolean> {
-  try {
     const hermesEnabled = await parseGradlePropertiesFile(gradleFile);
     if (hermesEnabled !== undefined) {
       return hermesEnabled;
@@ -507,10 +506,6 @@ export async function getAndroidHermesEnabled(gradleFile: string): Promise<boole
     return parseBuildGradleFile(gradleFile).then((buildGradle: any) => {
       return Array.from(buildGradle["project.ext.react"] || []).some((line: string) => /^enableHermes\s{0,}:\s{0,}true/.test(line));
     });
-  } catch (error) {
-    console.error(error);
-    return false;
-  }
 }
 
 async function getHermesCommandFromGradle(gradleFile: string): Promise<string> {
--- a/src/commands/codepush/lib/react-native-utils.ts
+++ b/src/commands/codepush/lib/react-native-utils.ts
@@ -485,10 +485,21 @@ async function getHermesCommandFromGradle(gradleFile: string): Promise<string> {
   }
 }
 
-export function getAndroidHermesEnabled(gradleFile: string): boolean {
-  return parseBuildGradleFile(gradleFile).then((buildGradle: any) => {
-    return Array.from(buildGradle["project.ext.react"] || []).some((line: string) => /^enableHermes\s{0,}:\s{0,}true/.test(line));
-  });
+export async function getAndroidHermesEnabled(gradleFile: string): Promise<boolean> {
+  // find hermesEnabled in gradle.properties first
+  try {
+    const properties = await g2js.parseFile(path.join("android", "gradle.properties"));
+    if (typeof properties["hermesEnabled"] !== "undefined") {
+      return properties["hermesEnabled"];
+    }
+    const enabled = !!(await parseBuildGradleFile(gradleFile).then((buildGradle: any) => {
+      return Array.from(buildGradle["project.ext.react"] || []).some((line: string) => /^enableHermes\s{0,}:\s{0,}true/.test(line));
+    }));
+    return enabled;
+  } catch (error) {
+    out.text(error);
+    return false;
+  }
 }
 
 export function getiOSHermesEnabled(podFile: string): boolean {
--- a/src/commands/codepush/lib/react-native-utils.ts
+++ b/src/commands/codepush/lib/react-native-utils.ts
@@ -455,7 +455,7 @@ export async function runHermesEmitBinaryCommand(
   });
 }
 
-function parseBuildGradleFile(gradleFile: string) {
+function getBuildGradleFilePath(gradleFile: string) {
   let buildGradlePath: string = path.join("android", "app");
   if (gradleFile) {
     buildGradlePath = gradleFile;
@@ -468,9 +468,49 @@ function parseBuildGradleFile(gradleFile: string) {
     throw new Error(`Unable to find gradle file "${buildGradlePath}".`);
   }
 
+  return buildGradlePath;
+}
+
+function parseBuildGradleFile(gradleFile: string) {
+  const buildGradlePath: string = getBuildGradleFilePath(gradleFile);
   return g2js.parseFile(buildGradlePath).catch(() => {
     throw new Error(`Unable to parse the "${buildGradlePath}" file. Please ensure it is a well-formed Gradle file.`);
   });
+}
+
+// parse gradle.properties file
+async function parseGradlePropertiesFile(gradleFile: string) {
+  // find gradle.properties file based on gradleFile
+  const buildGradlePath: string = getBuildGradleFilePath(gradleFile);
+  const gradlePropertiesPath: string = path.join(path.dirname(buildGradlePath), "../", "gradle.properties");
+  // check if gradle.properties file exists
+  if (fileDoesNotExistOrIsDirectory(gradlePropertiesPath)) {
+    console.error(`Unable to find gradle.properties file at "${gradlePropertiesPath}".`);
+    return;
+  }
+  try {
+    const properties = await g2js.parseFile(gradlePropertiesPath);
+    if (properties && typeof properties["hermesEnabled"] === "boolean") {
+      return properties["hermesEnabled"];
+    }
+  } catch (error) {
+    console.error(`Unable to parse the "${gradlePropertiesPath}" file. Please ensure it is a well-formed Gradle file.`);
+  }
+}
+
+export async function getAndroidHermesEnabled(gradleFile: string): Promise<boolean> {
+  try {
+    const hermesEnabled = await parseGradlePropertiesFile(gradleFile);
+    if (hermesEnabled !== undefined) {
+      return hermesEnabled;
+    }
+    return parseBuildGradleFile(gradleFile).then((buildGradle: any) => {
+      return Array.from(buildGradle["project.ext.react"] || []).some((line: string) => /^enableHermes\s{0,}:\s{0,}true/.test(line));
+    });
+  } catch (error) {
+    console.error(error);
+    return false;
+  }
 }
 
 async function getHermesCommandFromGradle(gradleFile: string): Promise<string> {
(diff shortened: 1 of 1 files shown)


BANG88:

src/commands/codepush/lib/react-native-utils.ts
```diff
@@ -485,10 +485,21 @@ async function getHermesCommandFromGradle(gradleFile: string): Promise<string> {
   }
 }
 
-export function getAndroidHermesEnabled(gradleFile: string): boolean {
-  return parseBuildGradleFile(gradleFile).then((buildGradle: any) => {
-    return Array.from(buildGradle["project.ext.react"] || []).some((line: string) => /^enableHermes\s{0,}:\s{0,}true/.test(line));
-  });
+export async function getAndroidHermesEnabled(gradleFile: string): Promise<boolean> {
+  // find hermesEnabled in gradle.properties first
```
DmitriyKirakosyan, line 489: I guess the comment is misplaced. Let's put it right under the parsing `gradle.properties`.

src/commands/codepush/lib/react-native-utils.ts
```diff
@@ -485,10 +485,21 @@ async function getHermesCommandFromGradle(gradleFile: string): Promise<string> {
...
+  try {
+    const properties = await g2js.parseFile(path.join("android", "gradle.properties"));
```
DmitriyKirakosyan, line 491: what happens if the file doesn't not exist? Will it crash? If yes, then it won't try to parse `gradleFile`, which is not what we would want.

src/commands/codepush/lib/react-native-utils.ts
```diff
@@ -485,10 +485,21 @@ async function getHermesCommandFromGradle(gradleFile: string): Promise<string> {
...
+    if (typeof properties["hermesEnabled"] !== "undefined") {
```
DmitriyKirakosyan, line 492: again, I guess it can throw exception is properties is not an object. Is there only "undefined" value possible? Can it be null or anything like that? May be worth adding tests.

src/commands/codepush/lib/react-native-utils.ts
```diff
@@ -485,10 +485,21 @@ async function getHermesCommandFromGradle(gradleFile: string): Promise<string> {
...
+      return properties["hermesEnabled"];
```
DmitriyKirakosyan, line 493: I think even if property exists, we would like to make sure it is a boolean.

src/commands/codepush/lib/react-native-utils.ts
```diff
@@ -485,10 +485,21 @@ async function getHermesCommandFromGradle(gradleFile: string): Promise<string> {
...
+    }
+    const enabled = !!(await parseBuildGradleFile(gradleFile).then((buildGradle: any) => {
```
DmitriyKirakosyan, line 495: Why extra var? just return. Why `!!`?

src/commands/codepush/lib/react-native-utils.ts
```diff
@@ -485,10 +485,21 @@ async function getHermesCommandFromGradle(gradleFile: string): Promise<string> {
...
+      return Array.from(buildGradle["project.ext.react"] || []).some((line: string) => /^enableHermes\s{0,}:\s{0,}true/.test(line));
+    }));
+    return enabled;
+  } catch (error) {
+    out.text(error);
```
DmitriyKirakosyan, line 500: It's better to use `console.error` here.

src/commands/codepush/lib/react-native-utils.ts
BANG88: Sure.

BANG88:
I have updated it. Thank you.

src/commands/codepush/lib/react-native-utils.ts
```diff
@@ -468,11 +468,51 @@ function parseBuildGradleFile(gradleFile: string) {
...
+  const buildGradlePath: string = getBuildGradleFilePath(gradleFile);
   return g2js.parseFile(buildGradlePath).catch(() => {
     throw new Error(`Unable to parse the "${buildGradlePath}" file. Please ensure it is a well-formed Gradle file.`);
   });
 }
 
+// parse gradle.properties file
+async function parseGradlePropertiesFile(gradleFile: string) {
+  // find gradle.properties file based on gradleFile
+  const buildGradlePath: string = getBuildGradleFilePath(gradleFile);
```
MikhailSuendukov, line 484: I think it's better to remove the dependence of this function on the function of getting the path to the build gradle file.

src/commands/codepush/lib/react-native-utils.ts
```diff
@@ -468,11 +468,51 @@ function parseBuildGradleFile(gradleFile: string) {
...
+  const gradlePropertiesPath: string = path.join(path.dirname(buildGradlePath), "../", "gradle.properties");
+  // check if gradle.properties file exists
+  if (fileDoesNotExistOrIsDirectory(gradlePropertiesPath)) {
+    console.error(`Unable to find gradle.properties file at "${gradlePropertiesPath}".`);
+    return;
```
MikhailSuendukov, line 489: In this case, it is better to write that we return null

src/commands/codepush/lib/react-native-utils.ts
```diff
@@ -468,11 +468,51 @@ function parseBuildGradleFile(gradleFile: string) {
...
+export async function getAndroidHermesEnabled(gradleFile: string): Promise<boolean> {
+  try {
+    const hermesEnabled = await parseGradlePropertiesFile(gradleFile);
+    if (hermesEnabled !== undefined) {
+      return hermesEnabled;
+    }
+    return parseBuildGradleFile(gradleFile).then((buildGradle: any) => {
+      return Array.from(buildGradle["project.ext.react"] || []).some((line: string) => /^enableHermes\s{0,}:\s{0,}true/.test(line));
+    });
+  } catch (error) {
```
MikhailSuendukov, line 510: In the previous implementation, there was no error trapping here. I think it's better to leave the old behavior

MikhailSuendukov (changes requested):

src/commands/codepush/lib/react-native-utils.ts
BANG88: Could you explain why we need this? thank you.

src/commands/codepush/lib/react-native-utils.ts
BANG88: Okey. ` return undefined ` now.

src/commands/codepush/lib/react-native-utils.ts
MikhailSuendukov: Because I think this dependency is redundant. We can safely get the path to the property file without using the path to the build.gradle file, using only the argument of this function.

BANG88: @DmitriyKirakosyan Sorry I can't understand what should I do. 😢

MikhailSuendukov: Sorry, it looks like I misunderstood part of the code, so you can ignore this comment of mine.

src/commands/codepush/lib/react-native-utils.ts
```diff
@@ -468,11 +468,46 @@ function parseBuildGradleFile(gradleFile: string) {
...
+    return;
+  }
+  try {
+    const properties = await g2js.parseFile(gradlePropertiesPath);
+    if (properties && typeof properties["hermesEnabled"] === "boolean") {
+      return properties["hermesEnabled"];
+    }
+  } catch (error) {
+    console.error(`Unable to parse the "${gradlePropertiesPath}" file. Please ensure it is a well-formed Gradle file.`);
+  }
```
DmitriyKirakosyan, line 498: Function which returns a value should always return. Let's put another `return undefined` at the end.

src/commands/codepush/lib/react-native-utils.ts
```diff
@@ -468,11 +468,46 @@ function parseBuildGradleFile(gradleFile: string) {
     throw new Error(`Unable to find gradle file "${buildGradlePath}".`);
```
DmitriyKirakosyan, line 468: I would put the check `if (fileDoesNotExistOrIsDirectory(buildGradlePath)) {` in `parseBuildGradleFile` instead, because I feel existence of this file shouldn't affect whether we try parsing the gradle properties.

src/commands/codepush/lib/react-native-utils.ts
```diff
@@ -455,7 +455,7 @@ export async function runHermesEmitBinaryCommand(
   });
 }
 
-function parseBuildGradleFile(gradleFile: string) {
+function getBuildGradleFilePath(gradleFile: string) {
```
DmitriyKirakosyan, line 458: let's rename these two methods to `getGradleBuildFilePath` and `parseGradleBuildFile`, for the sake of consistency with `parseGradlePropertiesFile`.

Also I would put `hermes` somehow in this names of the parsing methods, because this file is `react-native-utils` and function names such as `parseGradleBuildFile` don't reflect the logic inside. It should be more specific to make it clear. F.e. `checkHermesInGradlePropertiesFile` and `checkHermesInGradleBuildFile`.

src/commands/codepush/lib/react-native-utils.ts
```diff
@@ -468,11 +468,46 @@ function parseBuildGradleFile(gradleFile: string) {
...
+}
+
+function parseBuildGradleFile(gradleFile: string) {
+  const buildGradlePath: string = getBuildGradleFilePath(gradleFile);
   return g2js.parseFile(buildGradlePath).catch(() => {
     throw new Error(`Unable to parse the "${buildGradlePath}" file. Please ensure it is a well-formed Gradle file.`);
   });
 }
 
+// parse gradle.properties file
```
DmitriyKirakosyan, line 481: Let's remove this comment and the two below. I don't think we need a comment for the clear code, which is easily understandable when you read it.

src/commands/codepush/lib/react-native-utils.ts
```diff
@@ -468,11 +468,46 @@ function parseBuildGradleFile(gradleFile: string) {
     throw new Error(`Unable to find gradle file "${buildGradlePath}".`);
   }
 
+  return buildGradlePath;
+}
+
+function parseBuildGradleFile(gradleFile: string) {
+  const buildGradlePath: string = getBuildGradleFilePath(gradleFile);
   return g2js.parseFile(buildGradlePath).catch(() => {
```
DmitriyKirakosyan, line 476: why don't we await here? I think the behaviour should be the same as in `parseGradlePropertiesFile`.

src/commands/codepush/lib/react-native-utils.ts
BANG88: Okay.

src/commands/codepush/lib/react-native-utils.ts
BANG88: Leave it as it is. 😄

src/commands/codepush/lib/react-native-utils.ts
MikhailSuendukov: @BANG88 Could you please rename the name of this function in the way Dmitry described?

BANG88: Okay. but  rename which one ? thanks.

MikhailSuendukov: Firstly, you have an error in the names of the functions, you have two functions, one of which is responsible for getting the path to the gradle build file and the second for parsing this file, but nevertheless they are both called the same `getBuildGradleFilePath`. You need to rename the function that is responsible for getting the path to the gradle build file to `getGradleBuildFilePath`. And the one that is responsible for parsing on `checkHermesInGradleBuildFile`. Also rename the function that is responsible for parsing the gradle property file to `checkHermesInGradlePropertiesFile`, because the parsing functions refer specifically to the logic of finding the attribute of the included Hermes engine.

BANG88: I'm sorry, I really am too stupid, blind.

MikhailSuendukov:
Firstly, the checkHermesInGradlePropertiesFile function does not work with the properties file, but with the build file, and secondly, it does not check if Hermes is included in this file, but only parses it. Thirdly, the parseGradlePropertiesFile function not only parses the properties file, but also checks whether Hermes is enabled in this file. Please change the behavior and name of the functions so that there is a function for parsing the build file and parsing the properties file, and checking that Hermes is enabled is in the getAndroidHermesEnabled function.

BANG88:
Sorry.

MikhailSuendukov:
Thank you very much for your contribution. If you decide to continue working on this pull request, we will be glad to see you here again.

PR #2341, opened 2023-06-05
Allow deployment name to be a number

adrienharnay:
Allow deployment name to be a number when invoking `appcenter codepush deployment add/remove`


adrienharnay:
@microsoft-github-policy-service agree company="Brigad"

lucen-ms (approved):

